PR #29, opened 2019-04-05
Ignore null in Graql Aggregate Hashcode

flyingsilverfin:
Fix null pointer error when trying to hash the GraqlGet - Aggregate subtype.

Ignore the `var` field if it is null


haikalpribadi:
why was this added to milestone 1.0 @flyingsilverfin ? 1.0 has been released last week. cc: @lolski

lolski:
NOTE: this PR is a follow-up to https://github.com/graknlabs/graql/pull/29

PR #30, opened 2019-04-05
Fix aggregate hashcodes

flyingsilverfin:
Fix NullPointerException in `hashCode` with GraqlGet - Aggregate

Fix two places where `var` could be null and ignore these nulls in the `hashCode` method


haikalpribadi:
this also shouldn't be in milestone 1.0 @flyingsilverfin

PR #54, opened 2019-04-15
Rename to asRetrievedVar() and isRetrieved

### LINKED ISSUE #53: Rename `asUserDefined`
The naming for `asUserDefined` variables is very confusing - it actually means whether or not to retrieve the variable as part of the answer.

Propose renaming `var.asUserDefined()` and `var.isUserDefinedName()` to `var.asRetrievedVar()` and `var.isRetrieved()`.

flyingsilverfin:
Clearer naming when indicating a variable is to be returned as part of the `get` clause. This is because in many places `asUserDefined` is not actually a user defined variable (especially in reasoner and QP) but instead used to ensure the variable is part of the `get` clause. This has lead to confusion about whether it's really about being able to check whether a variable was originally created by the user, or whether it's means to be returned in the `get`. The two main known use cases are:
1. Users creating variables with `var()`, then setting `asUserDefined()` to avoid creating a name for a variable they want to see in the `get()`
2. Internally, explicitly wanting to have the variable be returned.
Both boil down to creating anonymous variables that should be returned.

Rename `asUserDefined()` to `asReturnedVar()` and `isUserDefinedName()` to `isReturned()`. This closes #53.

### DIFF
--- a/java/property/HasAttributeProperty.java
+++ b/java/property/HasAttributeProperty.java
@@ -81,7 +81,7 @@ public class HasAttributeProperty extends VarProperty {
 
         property.add(type);
 
-        if (attribute().var().isUserDefinedName()) {
+        if (attribute().var().isRetrieved()) {
             property.add(attribute().var().toString());
         } else {
             attribute().getProperties(ValueProperty.class).forEach(prop -> property.add(prop.operation().toString()));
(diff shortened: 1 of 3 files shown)


haikalpribadi:
I don't quite understand this fully, yet, @flyingsilverfin. Are all variables that have `isUserDefined = true` actually are defined by the user? The fact that it will eventually be retrieved - isn't that implementation detail?

flyingsilverfin:
@haikalpribadi internally (including in reasoner and in the QP), the main purpose of `isUserDefined` is actually to ensure that the variable is retrieved as part of the `get`, so the naming doesn't make sense for this purpose. And as far as I'm aware, users essentially use it for the same purpose ie. creating an anonymous variable that is then labeled as user defined so that it is included in the answer without having to declare an explicit variable name. So it feels like the purpose here is to make sure the variable is retrieved rather than a flag of whether the user created it.

PR #80, opened 2019-09-23
Use Github credential for release-cleanup

lolski:

Use GitHub credential for release-cleanup. This removes the need for adding an additional SSH key to CircleCI.


haikalpribadi:
Don't forget to delete the old SSH Key @lolski

PR #94, opened 2019-12-09
Datetime Milliseconds Decimal Representation

### LINKED ISSUE #93: Ability to store milliseconds in date-time attribute
## Problem to Solve

At the moment, we are not able to store date-time value at the granularity of Milliseconds. Our date-time format only allows us to save time data down to seconds: `yyyy-MM-ddTHH:mm:ss`.

## Current Workaround

One would have to store the milliseconds' data in a separate attribute of type `long`.

## Proposed Solution

Our internal representation of date-time attributes are already a `long` value of time since Epoch in milliseconds. All we need to do now is to extend Graql's date-time data type to be able to represent milliseconds. The format of we can define it to be in Graql can be the following:
```
yyyy-MM-ddTHH:mm:ss.SSS
```

## Additional Information

This feature will need to be followed up with a change in Grakn Core (issue https://github.com/graknlabs/grakn/issues/5533)

flyingsilverfin:
Enable and test the inclusion of fractions of seconds in a date time attribute value, up to 3 digits, allowing us to store up to 999 milliseconds.
For example, `2019-01-01T10:10:10.123` will now parse as having `10.123` seconds, and `2019-01-01T10:10:10.1` will parse as having `10.100` seconds

Closes #93 

* Implement and Test up to 3 digits after a `.` after seconds in a datetime

### DIFF
--- a/grammar/Graql.g4
+++ b/grammar/Graql.g4
@@ -346,11 +346,11 @@ fragment DATE_FRAGMENT_ : YEAR_ '-' MONTH_ '-' DAY_ ;
 fragment MONTH_         : [0-1][0-9] ;
 fragment DAY_           : [0-3][0-9] ;
 fragment YEAR_          : [0-9][0-9][0-9][0-9] | ('+' | '-') [0-9]+ ;
-fragment TIME_          : HOUR_ ':' MINUTE_ (':' SECOND_ ('.' MILLIS_)? )? ;
+fragment TIME_          : HOUR_ ':' MINUTE_ (':' SECOND_ ('.' SECOND_FRACTION_)? )? ;
 fragment HOUR_          : [0-2][0-9] ;
 fragment MINUTE_        : [0-6][0-9] ;
 fragment SECOND_        : [0-6][0-9] ;
-fragment MILLIS_        : [0-9] ([0-9] ([0-9])?)?; // between 1 and 3 digits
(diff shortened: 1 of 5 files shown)


grammar/Graql.g4
```diff
@@ -346,10 +346,11 @@ fragment DATE_FRAGMENT_ : YEAR_ '-' MONTH_ '-' DAY_ ;
 fragment MONTH_         : [0-1][0-9] ;
 fragment DAY_           : [0-3][0-9] ;
 fragment YEAR_          : [0-9][0-9][0-9][0-9] | ('+' | '-') [0-9]+ ;
-fragment TIME_          : HOUR_ ':' MINUTE_ (':' SECOND_)? ;
+fragment TIME_          : HOUR_ ':' MINUTE_ (':' SECOND_ ('.' MILLIS_)? )? ;
 fragment HOUR_          : [0-2][0-9] ;
 fragment MINUTE_        : [0-6][0-9] ;
-fragment SECOND_        : [0-6][0-9] ('.' [0-9]+)? ;
+fragment SECOND_        : [0-6][0-9] ;
+fragment MILLIS_        : [0-9] ([0-9] ([0-9])?)?; // between 1 and 3 digits
```
haikalpribadi: Let's call this `SECOND_FRACTION_` @flyingsilverfin

PR #98, opened 2020-01-16
Patch @rules_antlr to get Maven artifacts via secure URLs

vmax:

Recently Maven Central started enforcing HTTPS when downloading artifacts ([ref](https://support.sonatype.com/hc/en-us/articles/360041287334)). This PR fixes building of Graql by replacing artifact URLs with secure ones.

Note: in future, this should be addressed by upgrading `rules_antlr` but API has changed between the version we use and upstream.

- Fetch `@rules_antlr` with `http_archive` rule (supports applying patch)
- Supply `maven-https.patch` that replaces URLs by which ANTLR Java dependencies are obtained
- Add `.bazelversion` to signal correct version of Bazel to build with


lolski:
@vmax great work in finding the issue! Let's wait for Joshua to try upgrading `rules_antlr` to the latest one because that would be the more preferrable solution.

vmax:
Will replaced with forked version.

PR #102, opened 2020-02-12
JetBrains Graql Plugin Implementation

BFergerson:
Implementation of the JetBrains Graql plugin.

This project uses Gradle to build. I'm not sure if you guys want that but I don't know how Bazel works so I would need help doing the conversion (if the conversion is possible).

No tests yet but this is something that could be easily added. This plugin is already better than the original plugin I made so I figured I would get the conversation started.

I added the `ErrorCharacter : . ;` line to the .g4 file because I was running into https://github.com/antlr/antlr4-intellij-adaptor/wiki/Getting-started#about-grammars. I don't know if it's the most appropriate thing to do but it fixes the issue.


flyingsilverfin:
Looks like a good amount of work went into this! Cheers :) I think i might need to discuss with you how feasible it might be to migrate from Gradle to Bazel (as it's our preferred build system... however JetBrains seems to provide a Gradle-based build environment)

BFergerson:
@flyingsilverfin, not too much work. Leveraged a good amount from the previous plugin and it doesn't work for all of the Graql syntax yet. Covers more than the previous plugin though so I thought I'd make a PR.

As for migrating to Bazel, I'm not sure how that would work. I make use of:
 - https://github.com/JetBrains/gradle-intellij-plugin
 - https://github.com/antlr/antlr4-intellij-adaptor

And I'm not sure those have Bazel equivalents. I think if you replaced them you would have to re-write those plugins from scratch or deal with hardcoding a lot of stuff. Not too sure.

flyingsilverfin:
Oh, also, some tests would by great, so that if the grammar changes we get a notification the plugin breaks :)

The gradle dependency from the intellij plugin dev environment is exactly the issue - we definitely dont want to rewrite their plugin. I will discuss with the team whether we can make the exception to use gradle here!

BFergerson:
@flyingsilverfin, added a Graql parsing test. This will ensure that the plugin's test breaks when the syntax changes in ways that would break the plugin.

Do you have a small snippet of code that shows all the syntax possibilities (or a large amount) of Graql? Right now I'm just using code I copied from Grakn documentation but I know it doesn't use everything Graql can do (like rules). So if syntax changes in something outside what's [here](https://github.com/graknlabs/graql/blob/56b5446fd6cac8e27a0df35bb64c6168c24202f5/plugins/jetbrains-plugin/src/test/resources/ParsingTestData.gql), it won't notice.

I would also like it for the syntax coloring screen:
Screenshot from 2020-02-13 04-13-39

flyingsilverfin:
James is building a very comprehensive schema here: https://github.com/graknlabs/simulation/blob/master/schema
there are two parts to this schema :) Does this help?

haikalpribadi:
This is amazing! Thanks, @BFergerson !

flyingsilverfin:
@BFergerson good news! We're happy to have your plugin built using gradle - we will aim to have plugin being published and tested in our build system (from/via bazel) in the future :) Sound good?

haikalpribadi:
I'll delegate the review of this PR to @flyingsilverfin :)

flyingsilverfin:
Note: to get this to work in IntelliJ while also using the Bazel plugin the Grakn team typically uses during the development workflow:
1. Open `graql` and bazel project
2. Create a gradle run configuration with Gradle project as `jetbrains-plugin` and task `
3. This should expand a gradle menu with now having the menu options such as `runIde`

flyingsilverfin:
@BFergerson super cool!!! little Graql icon is great :) 

One thing, I tried running it via `runIde` and i get things like this:

trying with schema:
```define
    person sub entity,
      plays test;
    employment sub relation,
      relates test;
    hello-world sub entity;
    name sub attribute, datatype long;
    person sub entity, has name;


insert
    $x isa person; $r (test: $x) isa employment;
```

leads to a red underline under `define` with error:
```mismatched input 'insert' expecting {<EOF>, 'thing', 'entity', 'attribute', 'relation', 'role', 'rule', 'count', 'max', 'min', 'mean', 'median', 'std', 'sum', 'cluster', 'path', 'degree', 'k-core', 'connected-component', 'from', 'to', 'of', 'in', 'where', 'min-k', 'k', 'size', 'contains', VAR_, ID_, TYPE_IMPLICIT_, TYPE_NAME_}```

Also it seems like the linking between defined and unused types isn't quite working? See screenshot below: `test` and `name` red underlining says "Concept ____ not defined", and then the yellow background when they are underlined says "unused". 

Screen Shot 2020-02-20 at 12 17 50 PM

**Comment by BFergerson** [comment]
@flyingsilverfin, thanks for taking a look. Not seeing either of those issues with the latest commit.

**Comment by haikalpribadi** [comment]
Can I approve and merge this now @flyingsilverfin ?

File: grammar/Graql.g4 [hunk 1]
```diff
@@ -355,3 +355,4 @@ fragment ESCAPE_SEQ_    : '\\' . ;
 
 COMMENT         : '#' .*? '\r'? ('\n' | EOF)    -> channel(HIDDEN) ;
 WS              : [ \t\r\n]+                    -> channel(HIDDEN) ;
+ErrorCharacter : . ;
```
haikalpribadi: Can we look into why this fixes the issue, and if there are any side effects to the Graql language, @BFergerson @flyingsilverfin ? And if it's the correct use, please update the PR description, as well as the alignment of the colon `:` to match the lines above it.

BFergerson: Fixed the alignment. I'm unsure what you mean about updating PR description. The PR description already mentions this change and why it was done.

plugins/jetbrains-plugin/build.gradle
```diff
@@ -0,0 +1,38 @@
+plugins {
```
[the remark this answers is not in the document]
BFergerson: Added basic readme with relevant links

grammar/Graql.g4
flyingsilverfin: I'll investigate why we have to have the `ErrorCharacter` and let you know :)
Sorry to be so picky on this, we're very protective of the Graql repository since it's one of the main value props of grakn! We'll get this merged soon!

BFergerson: No rush on my part. I'm already using this plugin :).

This is what I was running into: https://github.com/antlr/antlr4-intellij-adaptor/wiki/Getting-started#about-grammars

## Recreating
Make schema
```
define
    person sub entity;
```

Modify to:
```
define
    person sub entity;
```

Try to delete `@`.

flyingsilverfin:
@BFergerson `verification/test` is currently failing  :)

BFergerson:
@flyingsilverfin, fixed.

BFergerson:
Currently stuck on this: https://intellij-support.jetbrains.com/hc/en-us/community/posts/360007668140-Inspections-and-Go-To-Declaration-require-cut-paste-before-detecting-correct-structure

flyingsilverfin:
Wow straight to the source :) Thanks so much brandon!! Let me know if we can help somehow

flyingsilverfin:
Any updates on this @BFergerson ?

BFergerson:
@flyingsilverfin, sorry for the delay. I reached out to the IntelliJ plugin support developers on Slack about two weeks ago and someone said they would look at it but never got back to me. I've bumped my forum post so hopefully, they will take another look.

BFergerson:
@neomatrix369 asked about that status of this PR so I gave the issue I was facing above another swing. I was able to solve it by ensuring the correct type is applied whenever `subtreeChanged` is called. This probably isn't the most efficient approach but it seems to work as far as I've been testing it.

@flyingsilverfin, please take another look at this when you can.

flyingsilverfin:
will try today!

flyingsilverfin:
@BFergerson mind merging in from master? We updated our `Bazel` version everywhere

BFergerson:
@flyingsilverfin, merged.

flyingsilverfin:
Hey brandon, sorry the week's been super busy! I'm going to get @alexjpwalker to take a look at this first thing on Monday!

thanks :))

alexjpwalker:
@BFergerson hi there, I've had a look at the plugin which looks great, being able to autocomplete keywords such as relates and value is very nice indeed and will be a useful tool for future Grakn developers.

I'm testing it against the following valid Graql file:

image

source:
```
define

person sub entity, plays test;
employment sub relation, relates test;
hello-world sub entity;
name sub attribute, value long;
person sub entity, has name;

insert
    $x isa person; $r (test: $x) isa employment;
```

It incorrectly shows:
line 3: concept 'test' has not been defined
line 4: role 'test' is never used
line 6: attribute 'name' is never used
line 7: concept 'name' has not been defined

In each case, the expected behaviour is "this syntax is valid".

BFergerson:
@alexjpwalker, thanks for taking a look. Make sure you're using the latest version I've committed because I felt pretty confident I solved that issue.

Also, I'm unable to recreate it:
Peek 2020-05-27 22-21

alexjpwalker:
> @alexjpwalker, thanks for taking a look. Make sure you're using the latest version I've committed because I felt pretty confident I solved that issue.
@BFergerson
Should be up-to-date, yes. Here are the last 5 commits in my `git log`:

image

which match the last 5 commits in this PR.

Maybe you have uncommitted code, an unmerged branch or a new Gradle configuration? I am using the Gradle task `runIde` to playtest the plugin.

BFergerson:
I have no uncommitted/pushed code on this. What I can tell you is that the sequence of keystrokes you use to produce this bug may be crucial in allowing me to reproduce it. I can remember a scenario where a bug like what you're experiencing would only occur after I've typed things incorrectly, deleted them, and then typed them correctly. As you can see with the gif I sent in the last comment, no issue occurs when things are typed correctly. I think I'll need you to provide a gif producing the bug for me to be able to fix.

alexjpwalker:
Took a couple of attempts but I managed to make a recording with no typos!

Screen Recording 2020-05-28 at 14 09 10

BFergerson:
@alexjpwalker, thanks for the gif. I was able to reproduce the issue. The issue you were facing didn't have anything to do with how the query was typed in. You could have pasted it in and got the same result. The issue was due to you using a scratch file. I don't commonly use scratch files so I never made sure it worked in them.

It should work correctly with the latest commit. Whenever you use a scratch file it'll consider the scratch file independently from the project, as it should.

alexjpwalker:
That's great - it now works as expected on my side!

I've noticed another thing but it is fairly minor - attaching a gif to show it:

Screen Recording 2020-05-28 at 17 31 45

(The schema was copy-pasted in; I expected autocomplete to bring up 'has', 'key', 'plays' etc. but actually it brought up "no suggestions")

In any case this doesn't seem like a blocker so I'm happy with the plugin. Assigning back to @flyingsilverfin for further review.

BFergerson:
Completion contribution isn't in the current plugin on the JetBrains Marketplace so it's not a blocker. Either way, I've added basic support. I didn't test it much but it should be correct.

haikalpribadi:
Why is this PR not being merged @flyingsilverfin @alexjpwalker ? Is there anything left to do / wait for? I worry that this will drag on to require more work with the 2.0 changes going through.

alexjpwalker:
I am very satisfied with the plugin in its current form.

haikalpribadi:
Good that we've now reviewed this PR to satisfaction, @alexjpwalker, @flyingsilverfin. Thank you for baring with us, @BFergerson ! We'll do a better job and reviewing PRs faster to get them merged in the future, Brandon!

For this specific instance, given Graql has now evolved to have the new 2.0 features in the `master` branch (there isn't many changes from the grammar point of view), by any chance we could upgrade this PR to work with the Graql 2.0 grammar? It has "scoped roles" which improves schema design and development significantly! We're yet to release a Grakn Core backend to execute the queries with this new grammar, but I presume you won't need that to just get JetBrains plugin to work?

grammar/Graql.g4
haikalpribadi: The content on the linked page above doesn't say anything else other than not to `-> skip`, @BFergerson - which we're not doing. It doesn't say why we need `ErrorCharacter: . ;` though. Did we ever find out why we need this character?

BFergerson: @haikalpribadi, given the schema: `define person@ sub entity;` what should the grammar file recognize `@` as?
I'm no expert on this stuff but I see it parsed as:
image

Alternatively, when I add `ErrorCharacter: . ;` it is parsed as:
image

The IntelliJ [documentation](https://jetbrains.org/intellij/sdk/docs/reference_guide/custom_language_support/implementing_lexer.html) says:
>  Lexers, and in particular JFlex-based lexers, need to be created in such a way that they always match the entire contents of the file, without any gaps between tokens, and generate special tokens for characters which are not valid at their location. Lexers must never abort prematurely because of an invalid character.

I don't know how much that applies but from my understanding the issue stems from the fact that the grammar without `ErrorCharacter` fails to recognize the `@` whereas adding the `ErrorCharacter` adds a catchall for anything that the grammar doesn't pick up, making sure every character is matched even when they're erroneous.

As the ANTLR plugin creator suggests, I copied the catchall from here: https://github.com/tunnelvisionlabs/antlr4-grammar-postgresql/blob/4c4d8e6a8f3057b319365ec947e6e76cc5a22e14/src/com/tunnelvisionlabs/postgresql/PostgreSqlLexer.g4#L808

BFergerson:
@haikalpribadi, I think I would need to see documentation on what queries should look like now. I don't know how to write the new queries so I wouldn't know how to tell if they're correct. I don't think there is much work involved though given that most of the logic is generated directly off the .g4 file.

grammar/Graql.g4
haikalpribadi: Thank you for the thorough explanation, @BFergeson! I completely understand what it's for now, and it makes complete sense! In fact, I should add this now in our current Grammar (as `UNRECOGNISED : . ;`) so it can be in sooner. You can remove it from your PR. Thanks for highlighting this, Brandon!

haikalpribadi: I've merged my commit from `master` actually, so you don't need to make any changes for this. I'll mark this issue resolved now.

haikalpribadi:
You're right I haven't written any docs for the new Graql syntax yet - I should have done a better job in the PR. But I can summarise it here as there are that many changes. First, we can see that there are 3 changes to the Grammar file that are relevant to this conversation. We can see them here: https://github.com/graknlabs/graql/commits/master/grammar/Graql.g4

---

1) [Graql 2.0 (#148)](https://github.com/graknlabs/graql/commit/8a4fcb24ca13e232fa2fa6842e0a016d42a5b1ef#diff-83d9cd31e39771ef1569c6c85c7fa6c3)

The first of two things this PR changes, is how the grammar parses semicolons, but they are still to be written the same way by the user. Semicolons previously were owned by the rule that it superseded. For example, `statement_thing` always need to end with `;`. However, now, `;` is no longer part of `statement_thing`, but expected to be provided to `delimit` multiple `statement_thing` when given in repetition. So effectively no change for the user.

Second, `statements` are now called `variables`. But they are effectively the same to the user.

---
2) [Removed all INTEGER, FLOAT, and REAL, with just LONG and DOUBLE](https://github.com/graknlabs/graql/commit/c95110c71418d740186e6333c11ad08808317549#diff-83d9cd31e39771ef1569c6c85c7fa6c3)

In the context of the grammar, we simply renamed the tokens `INTEGER_` and `REAL_` to `LONG_` and `DOUBLE_`.

---

3) [New Graql 2.0 Features (#150)](https://github.com/graknlabs/graql/commit/2fb1f5184829048c4e9d909b9f3dfd212410f844)

This is the PR that actually contains changes to the syntax, and contains new features.

a) Concept IDs:
- Rename `id` to `iid`
- Changed regex of concept IIDs from: `('V'|'E')[a-z0-9-]* ;` (alpha numeric string starting with `V` or `E`) to `'0x'[0-9a-f]+` (a hexadecimal string).

Example: Instead of saying `match $x id V123; get $x` you now say `match $x iid 0x04fa523; get $x`. 

Note how the token `id` is now freed u for the user to use.

b) Attribute ownership in the schema
- Replaced `has` keyword in Type Variable properties, with `owns`
- Replace `key` keyword with a `@key` annotation that follows `has <type>`

Example: Instead of writing
```
define person sub entity,
key email,
has name;
```
You now write:
```
define person sub entity,
owns email @key,
owns name;
```

d) Roles are now scoped by their relations, such that they will only be uniquely identifiable by `relation:role`.

Previously, names of Role Types are globally unique. Now, that's no longer true. Role Types only need to be unique in the scope of their relation, and the relation they extend/inherit. This simplifies the effort of naming role types. Thus, in order to refer to roles that a `thing` could `play`, we now refer to played roles as `relation:role`.

For example, previously we would say:
```
define

task sub entity,
    plays task-dependency_dependee,
    plays task-dependency_dependant;

task-dependency sub relation,
    relates task-dependency_dependee,
    relates task-dependency_dependant;
```

You can now say:
```
define

task sub entity,
    plays task-dependency:dependee,
    plays task-dependency:dependant;

task-dependency sub relation,
    relates dependee,
    relates dependant;
```

e) Overriding attribute and role types.
- override attribute types by defining `has <type> as <type>`
- override played role types by defining `plays <type> as <type>`
- override related role types by defining `relates <type> as <type>`

This is new. And it does not clash with previous grammar. Previously, we were not able to override an inherited attribute or role type. Now we can:
```
define 
company sub entity, has company-name;
charity sub company, has charity-name as company-name;

person sub entity, plays parentship:parent;
man sub person, plays fathership:father as parent;

parentship sub relation, relates parent;
fathership sub parentship, relates father as parent;
```

---

That's all the changes. I hope that helps, @BFergerson ! Would this be able to help you proceed?

PS: @flyingsilverfin @alexjpwalker @jmsfltchr we should my writing above for future reference. I'll add it to the previous PR.

haikalpribadi:
Another change that went into `master` last week, @BFergerson:

We've replaced "Get Query" with "Match Query". `get ...;` is now merely a filter on 
`match ...;` no longer `needs` to be ended with `get;` all the time. Writing `match $x isa thing, has name $n;` is now already a "match query", in which you can optionally filter by saying `get $n;`.

BFergerson:
@haikalpribadi, I haven't forgotten about this and I plan to work on it soon. I was invited to speak at [ApacheCon](https://www.apachecon.com/acah2020/tracks/observability.html) so all my time has been going towards trying to perfect my presentation. I will have time to work on this next month.

p.s. You guys are partially to blame for getting me invited to speak at another conference. The fact that I don't have time to work on this is something you did to yourself :P

haikalpribadi:
Hahaha but that's amazing, @BFergerson !!!! Massive congrats! Please do share with us the video once it's out :)

haikalpribadi:
There's also a third change to the grammar terminology now: variable `property` is now called `constraint`. I've added this to the long description above.

flyingsilverfin:
@BFergerson  just updated haikal's big comment with changes to rule syntax, that I will write here as well
======================================
f) New rule syntax
Rule definition no longer follows `Concept Type` definition syntax using `sub`, but now uses explicit rule define and undefine syntax:
```
define
rule transitive-ownership: 
when {
  (owned: $x, owner: $y) isa ownership;
  (owned: $y, owner: $z) isa ownership;
} then {
  (owned: $x, owner: $z) isa ownership;
};
```

```
undefine
rule transitive-ownership;
```

haikalpribadi:
g) New concept equality constraint, and removal of concept inequality constraint

We would like to streamline the attribute value constraints to be more consistent. Right now, `>`, `<`, `>=`, and`<=`, all operate on _"attribute values"_. However, `!=` operates on _"concept equality"_. And thus to evaluate _"attribute value equality"_ and _"inequality"_, we must use `==` and `!==`. This is very inelegant and thus could be error-prone as they are very close `=` and `!=`. This behaviour is also inelegant for another reason: that it introduces a second way to do _"negation"_, which is not through our `not` logical operator. This issue has also be in raised through issue #167. 

Our conclusion is to remove _"concept inequality"_ constraint (`!=`), and introduce new constraint operator: `is`. `is` would allow us to assert that 2 concepts are equal: `$x is $y`, which we can then negate by doing `not { $x is $y };`.

h) Update operator for Attribute Value equality and inequality

Given the change above, we can now use `=` and `!=` that have been freed up, as _"attribute equality"_ and _"inequality"_.

BFergerson:
Closing in favor of https://github.com/typedb-osi/typeql-plugin-jetbrains/pull/1.

As far as I know, I've incorporated everything mentioned in this PR. I'm more than willing to fix any issues that come up. I'd really just like to push this thing through first. It's far superior to what's currently on the JetBrains marketplace (which still receives hundreds of downloads per month).

Issue #115, opened 2020-04-02
Graql Delete queries should accept statements


haikalpribadi:
## Problem to Solve

This issue actually addresses several issues.

1) The ability to remove `attributes` from its owners without deleting the `attribute` itself
2) The ability to remove `roleplayers` from a relation

## Current Workaround

1) In order to delete an `attribute` from its owners right now uses the `via` keyword.
    ```
    match $x isa person, has name $n via $h; delete $h;
    ```
    However, we will be removing the `via` keyword, as we will be deleting the implicit relation that exists to implement the `has` keyword. Instead, we will have native `has` data structure.

2) Right now, there is no workaround to remove a `roleplayer` from a `relation` and this is actually a long-standing issue: #39.

## Proposed Solution

Allow the delete query to operate on full `statements`, just like insert `statements`. For example, here's an insert query operating on a `statement`:
```
match $x isa person; insert $x has name "alice";
```
A delete query operating on a `statement` can therefore do:
```
match $x isa person, has name $n; delete $x has name $n;
```
This will result in symmetry of `insert` and `delete` queries.

The above example shows how we solve the first problem - removing `attributes` from their owners`. In order to delete `roleplayers` from their relation, you can write
```
match $r (role-a: $x, role-b: $x) isa relation; 
delete $r (role-a: $x);
```

Note that in the examples above, we are not deleting the concept instances represented by the variables themselves. In order to delete the concept instances, you must write:
```
match $x isa person, has name $n;
delete $x isa person; $n isa name;
```
This will delete all instances of `person` and `name` found in the matched `pattern`.

### Concretely, the following will be implemented:

* we only allow `match...delete` queries (`delete...` only clauses are disallowed in the grammar)
* variables in the `delete` clause must be a subset of the `match` clause
* allowed deletion statements will by the same set allowed by `insert`
* as with inserts, using NEQ, Predicates (ie. operations) and ID properties in the `delete` clause should throw
* To verify that the concepts found in the `match` clause satisfy the `delete` clause as well: this means we can satisfy inheritance contraints naturally and throw otherwise, for example in :
```
match $x isa thing;
delete $x isa person;
```
however, the following will never throw:
```
match $x isa person;
delete $x isa thing;
```
* Deletion with ID edge case:
```
match $x id V123;
delete $x isa thing;
```
should work, however:
    * willl throw if V123 happens to be  a type
    * _requires_ the `isa thing` clause to delete the instance
[...]

PR #118, opened 2020-04-16
Deletes accept statement_instance instead of variables

flyingsilverfin:
Implementing the first step of https://github.com/graknlabs/graql/issues/115, we change the grammar to accept `statement_instance` for `delete` clauses, not just variables. As a medium term consequence, we remove the ability to limit, sort or offset delete queries.

* Change `variables` to `statement_instance` in `delete` clauses
* Update tests to reflect not just using variables
* Add new test to ensure query throws if `delete` variable not present in the `match`
* Remove `Filterable` from delete queries

### DIFF
--- a/grammar/Graql.g4
+++ b/grammar/Graql.g4
@@ -39,7 +39,7 @@ query_undefine      :   UNDEFINE    statement_type+ ;
 query_insert        :   MATCH       pattern+    INSERT  statement_instance+
                     |                           INSERT  statement_instance+  ;
 
-query_delete        :   MATCH       pattern+    DELETE  variables   filters  ;  // GET QUERY followed by aggregate fn
+query_delete        :   MATCH       pattern+    DELETE  statement_instance+ filters  ;  // DELETE QUERY followed by aggregate fn
 query_get           :   MATCH       pattern+    GET     variables   filters  ;  // GET QUERY followed by group fn, and
                                                                                 // optionally, an aggregate fn
 query_compute       :   COMPUTE     compute_conditions  ;
--- a/grammar/Graql.g4
+++ b/grammar/Graql.g4
@@ -39,7 +39,7 @@ query_undefine      :   UNDEFINE    statement_type+ ;
 query_insert        :   MATCH       pattern+    INSERT  statement_instance+
                     |                           INSERT  statement_instance+  ;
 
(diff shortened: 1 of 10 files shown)


flyingsilverfin:
@haikalpribadi  ready for review :)

java/query/MatchClause.java
```diff
@@ -116,73 +116,40 @@ public MatchClause(Conjunction<Pattern> pattern) {
...
     /**
-     * @param vars a collection of variables to delete for each result of this match clause
+     * @param delete a collection of statements that indicate properties to delete
      * @return a delete query that will delete the given variables for each result of this match clause
      */
     @CheckReturnValue
-    public final GraqlDelete.Unfiltered delete(LinkedHashSet<Variable> vars) {
-        return new GraqlDelete.Unfiltered(this, vars);
+    public final GraqlDelete delete(Collection<? extends Statement> delete) {
+        return new GraqlDelete(this, new ArrayList<>(delete));
```
haikalpribadi, line 152: variable `delete` should be named `statements`

grammar/Graql.g4
```diff
@@ -39,7 +39,7 @@ query_undefine      :   UNDEFINE    statement_type+ ;
 query_insert        :   MATCH       pattern+    INSERT  statement_instance+
                     |                           INSERT  statement_instance+  ;
```
haikalpribadi, line 41: You should remove the new lines between insert and delete, and also make sure `;` are aligned.

java/query/MatchClause.java
flyingsilverfin: ha! i had this before lol ok will go back

java/query/GraqlDelete.java
```diff
@@ -19,108 +19,77 @@
...
-    @Override
-    public Optional<Long> offset(){
-        return this.offset < 0 ? Optional.empty() : Optional.of(this.offset);
-    }
-
-    @Override
-    public Optional<Long> limit() {
-        return limit < 0 ? Optional.empty() : Optional.of(limit);
+    public Set<Variable> vars() {
+        return statements.stream().flatMap(statement -> statement.variables().stream()).collect(Collectors.toSet());
```
haikalpribadi, line 76: I should have caught it during the review, but we shouldn't have kept `vars()` in `GraqlDelete`, @flyingsilverfin . Given that it now takes in "instance statements", it should only be holding onto those, not caring about `vars()` and the equality check should be based on `statements` too.

PR #120, opened 2020-04-27
Add two tests for parsing query and query list

flyingsilverfin:
Test `Graql.parse` and `Graql.parseList` with multiline queries.

* Add new test cases


flyingsilverfin:
These tests are already covered in `ParserTest`

PR #137, opened 2020-06-25
Add scoped types to grammar

flyingsilverfin:
In order to scope roles to relations, we need to have extended syntax for scoped and unscoped types.


java/statement/Label.java
```diff
@@ -0,0 +1,98 @@
...
+
+    /**
+     * @param value The string which potentially represents a Type
+     * @return The matching Type Label
+     */
+    @CheckReturnValue
+    public static Label of(String value, @Nullable String scope) {
+        return new Label(value, scope);
+    }
+
```
flyingsilverfin, line 56: We could include another `.of(String name)` that doesn't require passing a null scope sometimes

java/statement/Label.java
```diff
@@ -0,0 +1,98 @@
...
+    public String name() {
```
flyingsilverfin, line 57: Also question here whether this name() method should return the full name WITH scope, or the default non-scoped name

flyingsilverfin:
Closed due to pointing to wrong target branch

Issue #107, opened 2020-02-27
Roles should be scoped within Relations


flyingsilverfin:
## Problem to Solve
Roles are currenty globally scoped, which not only makes naming more difficult, but also increases complexity of Graql as we can share roles across unrelated parts of the type hierarchy.

## Proposed Solution
We will scope roles to belong to relations. In other words, roles will be scoped to a relation and not live independently. This will making naming much simpler, as roles will be interpreted within the context of the relation name the user is providing.

flyingsilverfin:
Further, we want to add the ability to override a role in a sub-relation using `as`, which will block the sub-relation from inheriting the role form the parent. Not writing anything should simply inherit the role that is related in the parent.

flyingsilverfin:
Corner cases:
* undefine - undefine a parent role will require checking any children inheriting the role are not using it
* undefine - undefine a parent role that is override with `as` in children should be blocked as well

haikalpribadi:
Do we want to add this to the current project, @flyingsilverfin ?

flyingsilverfin:
No, i think it's a bit further out

haikalpribadi:
There will also be a syntax change to Graql, @flyingsilverfin :

When you define roles in a `relation`, such as
```
marriage sub relation,
  relates husband,
  relates wife;
```

You will refer to the roles being played by an `entity`, such as:
```
man sub person,
  plays marriage:husband;
```

When a relation inherits a role from a parent relation, such as the role `child` below:
```
parenthood sub relation,
  relates parent,
  relates child;

motherhood sub parenthood,
  relates mother as parent;
  # the role 'child' is inherited
```

You must can only refer to the `child` role through the relation it was first declared:
```
person sub relation,
  plays parenthood:child;
```

haikalpribadi:
Can you aggregate all the comments (both of yours and mine) into a clear breakdown of changes included in this issue, in the issue description, @flyingsilverfin? All this should be one big change that goes in together.

flyingsilverfin:
@haikalpribadi  sure - do you think inheriting roles is a prerequisite? the change is going to be big already, might be worth breaking it down where we can and doing first inheritance, then making the syntax changes.

haikalpribadi:
Note that some of the changes will be in Graql, and some are in Grakn Core. You can break them into the respective smaller task and put the issue in the correct repo, if that helps you. Make sure to reference them in this issue description - have a checklist of complete ones would be very useful.

[...]

PR #140, opened 2020-06-26
Add labels and scoped roles to the grammar and parsers

flyingsilverfin:
To being changes decided in #107 , we modify the grammar to have `type_scoped` as well as the default `type` that is unscoped. To make this make sense at the object level, we create a new object called `Label` that we will soon remove from `core` and `client-java` that can be scoped and unscoped types - Labels are now used throughout Graql to represent types. Where possible, we provide helper methods that consume `String`, creating the required `Label` objects internally and passing the Labels as required.

* Create `Label`, a class holding the combination of scope + name. `scope` may be null
* The parser creates `Label` classes instead of strings from `label_ref`, `label_scoped`, and `label` rules
* `Compute` targets such as `of` and `in` require `Label` objects under the hood, though we provide convenience helpers that take strings and wrap them as `Label` where possible


java/statement/Label.java
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    /**
+     * @param name The string which potentially represents a Type
+     * @return The matching Type Label
+     */
+    @CheckReturnValue
+    public static Label of(String name, @Nullable String scope) {
+        return new Label(name, scope);
+    }
```
flyingsilverfin, line 55: We could include another `.of(String name)` that doesn't require passing a null scope sometimes

java/statement/Label.java
```diff
@@ -0,0 +1,98 @@
...
+
+    public String name() {
+        return name;
+    }
```
flyingsilverfin, line 59: Also question here whether this name() method should return the full name WITH scope, or the default non-scoped name

java/statement/Label.java
haikalpribadi: Yes, add `.of(String name)`. We should only use `@Nullable` when we *really* have to.

java/statement/Label.java
haikalpribadi: `name()` can return just the unscoped label, and we can add `scopedName()` and return WITH scope.

java/Graql.java
```diff
@@ -247,15 +248,20 @@ private static Statement hiddenVar() {
     }
 
     @CheckReturnValue
-    public static StatementType type(Graql.Token.Type type) {
-        return type(type.toString());
+    public static StatementType type(Label label) {
+        return hiddenVar().type(label);
```
haikalpribadi, line 252: Why are we replacing `.type(Graql.Token.Type type)` with `.type(Label label)`, Josh? This was supposed to be used strictly for "native" types, right? Also, there are already 2 methods taking in labels as Strings below; they should be enough, right? Do we need one that strictly takes in an argument of `Class<Label>`? The user will then have to manually construct a `Label` class instance, to pass in, right?

java/property/TypeProperty.java
```diff
@@ -27,17 +28,17 @@
...
             throw new NullPointerException("Null label");
         }
-        this.name = name;
+        this.label = label;
     }
 
-    public String name() {
-        return name;
+    public Label label() {
+        return label;
```
haikalpribadi, line 41: Have we considered not having to implement a `Label` class, and just have `TypeProperty` to have `label` and `scope`, where both are represented as Strings?

java/statement/builder/StatementTypeBuilder.java
```diff
@@ -46,72 +47,67 @@ default StatementType type(Graql.Token.Type type) {
...
         return type(AbstractProperty.get());
     }
 
     default StatementType sub(Graql.Token.Type type) {
-        return sub(type.toString());
+        return sub(type.toString(), null);
+    }
+
+    @CheckReturnValue
+    default StatementType sub(String type, @Nullable String scope) {
```
haikalpribadi, line 75: I think we don't need the annotation `@Nullable` here?

java/statement/builder/StatementTypeBuilder.java
```diff
@@ -46,72 +47,67 @@ default StatementType type(Graql.Token.Type type) {
...
-     */
     @CheckReturnValue
     default StatementType sub(Statement type) {
         return sub(new SubProperty(type));
     }
 
+    @CheckReturnValue
+    default StatementType sub(SubProperty property) {
+        return type(property);
+    }
```
haikalpribadi, line 92: Why do we suddenly need this? 🤔

java/statement/builder/StatementTypeBuilder.java
```diff
@@ -46,72 +47,67 @@ default StatementType type(Graql.Token.Type type) {
...
     }
 
-    /**
-     * @param type a concept type id that this variable must be a kind of, without looking at parent types
-     * @return this
-     */
     @CheckReturnValue
-    default StatementType subX(String type) {
-        return subX(Graql.type(type));
+    default StatementType subX(String type, @Nullable String scope) {
```
haikalpribadi, line 99: This also doesn't need `@Nullable`. The user of this class needs to decide which method to call if they have or not have `scope`.

java/statement/builder/StatementTypeBuilder.java
```diff
@@ -192,7 +191,8 @@ default StatementType relates(Statement type) {
      */
     @CheckReturnValue
     default StatementType relates(String roleType, @Nullable String superRoleType) {
```
haikalpribadi, line 193: Should we also overload this?

java/property/TypeProperty.java
flyingsilverfin: yeah we need the class for the auto-generated parser, unless we want the auto-generated parser to return a Pair sometimes!

java/statement/builder/StatementTypeBuilder.java
flyingsilverfin: right

java/statement/builder/StatementTypeBuilder.java
flyingsilverfin: yeah lets stick to that style, I had not overloaded it to force me to update all the users properly

flyingsilverfin: lets unify the other methods too

java/statement/builder/StatementTypeBuilder.java
flyingsilverfin: it existed before - just rearranged some of the methods

java/property/TypeProperty.java
haikalpribadi: Right

java/statement/builder/StatementTypeBuilder.java
haikalpribadi: ok

java/statement/builder/StatementTypeBuilder.java
haikalpribadi: and by 'unify' you mean overload other methods too, right? and remove `@Nullable` annotation?

java/statement/Label.java
flyingsilverfin: yep that's what I've got!

PR #143, opened 2020-07-03
Allow `plays` to be scoped or unscoped

flyingsilverfin:
Relax constraints on `plays` to be scoped or unscoped. This led to a change in the grammar where now `type` is either `type_unscoped` or `type_scoped`, and correspondingly `type_label` which corresponds to `type_label_unscoped` and `type_label_scoped`

* Allow `plays` to take a `scoped` or `unscoped` type 
* Refactor the structure of the grammar to remove `type_ref`. The default `type` is now either an scoped OR unscoped type.

### DIFF
--- a/grammar/Graql.g4
+++ b/grammar/Graql.g4
@@ -95,18 +95,18 @@ pattern_statement   :   statement_type
 
 // TYPE STATEMENTS =============================================================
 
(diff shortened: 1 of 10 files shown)


flyingsilverfin:
Wrong target branch

PR #160, opened 2020-10-05
Implement rules for Graql 2.0

flyingsilverfin:
Implement rules with our new syntax:
```
rule transitive-location: when { 
  $located (inner: $x, outer: $y) isa location;
  $located2 (inner: $y, outer: $z) isa location;
} then {
  (inner: $x, outer: $z) isa location;
}
```

This constitutes a new `Definable` Graql fragment. `GraqlDefine` and `GraqlUndefine` consume `Definables`, which include `TypeVariable` and `Rule`. Note that `Rule` is not a Pattern, but a new schema construct we define.
  

* Implement `Rule`
* Implement `Definable` interface, with subtypes `Rule` and `TypeVariable`
* Update parser and grammar to handle the newly defined rule syntax
* Rules only accept conjunctions (list of pattern) in the `when` and a `ThingVariable` in the `then`
* Cleanup: Add `final` arguments, remove stray old `when` and `then` sources that should not exist anymore


grammar/Graql.g4
```diff
@@ -33,13 +36,13 @@ query                 :   query_define    |   query_undefine
...
-query_define          :   DEFINE      variable_types ;
-query_undefine        :   UNDEFINE    variable_types ;
+query_define          :   DEFINE      definable+  ;
+query_undefine        :   UNDEFINE    definable+  ;
 
 query_insert          :   MATCH       patterns      INSERT  variable_things
                       |                             INSERT  variable_things     ;
 query_delete          :   MATCH       patterns      DELETE  variable_things     ;
-query_match           :   MATCH       patterns            ( filters )?          ;
+query_match           :   MATCH       patterns            ( filters )           ;
```
flyingsilverfin, line 45: note: this was removed because having a `?` here as well as on each optional filter caused Antlr warnings

java/parser/Parser.java
```diff
@@ -506,33 +540,51 @@ public Pattern visitPattern(final GraqlParser.PatternContext ctx) {
...
             final List<Pattern> nested = visitPatterns(patternsContext);
             if (nested.size() > 1) return new Conjunction<>(nested);
             else return nested.get(0);
         }).collect(toList());
 
-        // Simplify representation when there is only one alternative
-        if (patterns.size() == 1) {
-            return patterns.iterator().next();
-        }
+        // TODO do we want this?
```
flyingsilverfin, line 550: discussion point

flyingsilverfin:
TODO: check tests are still valid, and decide where empty `when` and `then` validation will happen

grammar/Graql.g4
haikalpribadi: Did it not break the parser?

flyingsilverfin: why would it? each sub-filter is optional so this `filters` is optional as well

grammar/Graql.g4
```diff
@@ -203,14 +214,16 @@ type                  :   label                         | VAR_          ;
...
 label_any             :   label_scoped  | label         ;
 label_scoped          :   LABEL_SCOPED_ ;
-label                 :   LABEL_        | type_native   | unreserved    ;
+label                 :   LABEL_        | schema_native | type_native   | unreserved    ;
 labels                :   label         | label_array   ;
 label_array           :   '[' label ( ',' label )* ']'  ;
 
 // LITERAL INPUT VALUES =======================================================
 
+schema_native         : RULE              ;
```
haikalpribadi, line 223: Indentation of `RULE`

java/common/exception/ErrorMessage.java
```diff
@@ -29,26 +29,28 @@
...
-            new ErrorMessage(13, "The relation variable has not been provided with role players.");
+            new ErrorMessage(14, "The relation variable has not been provided with role players.");
     public static final ErrorMessage INVALID_CONSTRAINT_DATETIME_PRECISION =
-            new ErrorMessage(14, "Attempted to assign DateTime value of '%s' which is more precise than 1 millisecond.");
+            new ErrorMessage(15, "Attempted to assign DateTime value of '%s' which is more precise than 1 millisecond.");
     public static final ErrorMessage INVALID_DEFINE_QUERY_VARIABLE =
-            new ErrorMessage(15, "Invalid define/undefine query. User defined variables are not accepted in define/undefine query.");
+            new ErrorMessage(16, "Invalid define/undefine query. User defined variables are not accepted in define/undefine query.");
     public static final ErrorMessage REDUNDANT_NESTED_NEGATION =
             new ErrorMessage(16, "Invalid query containing redundant nested negations.");
```
haikalpribadi, line 55: There are 2 error messages with the code `16`.

java/parser/Parser.java
haikalpribadi: Replace with an assertion that `patterns.size()` is always larger than 1.

java/query/GraqlDefinable.java
```diff
@@ -30,43 +33,52 @@
...
-        this.variables = variables;
-        this.hash = Objects.hash(this.keyword, this.variables);
+        this.hash = Objects.hash(this.keyword, this.typeVariables, this.rules);
     }
 
-    public final List<TypeVariable> variables() {
-        return variables;
+    public final List<TypeVariable> typeVariables() {
+        return typeVariables;
+    }
```
haikalpribadi, line 67: I think this method can simply be called `variables()` as the context is clear: it is in `GraqlDefinable`. The only type of variable that can exist here is only `TypeVariable`. Same goes for `GraqlWritable` - only `ThingVariable` exists there, so that's what their `variables()` will return. We can revert the name of the class field too.

java/query/GraqlDefinable.java
```diff
@@ -30,43 +33,52 @@
...
+            if (definable.isRule()) rules.add(definable.asRule());
+            if (definable.isTypeVariable()) typeVars.add(definable.asTypeVariable());
+        }
+        while (!typeVars.isEmpty()) {
+            final TypeVariable v = typeVars.removeFirst();
             if (!v.isLabelled()) throw GraqlException.of(ErrorMessage.INVALID_DEFINE_QUERY_VARIABLE.message());
-            else v.constraints().forEach(c -> list.addAll(c.variables()));
+            else v.constraints().forEach(c -> typeVars.addAll(c.variables()));
+            typeVariables.add(v);
         }
```
haikalpribadi, line 59: Isn't this doable in 1 loop?

haikalpribadi: You just need to wrap the original loop's body with an `if` block to check if it's a variable, right?

java/query/GraqlDefinable.java
```diff
@@ -30,43 +33,52 @@
...
         final StringBuilder query = new StringBuilder();
         query.append(keyword);
 
-        if (variables.size() > 1) query.append(NEW_LINE);
+        if (typeVariables.size() + rules.size() > 1) query.append(NEW_LINE);
         else query.append(GraqlToken.Char.SPACE);
 
-        query.append(variables().stream().map(TypeVariable::toString).collect(joining("" + SEMICOLON + NEW_LINE)));
+        query.append(typeVariables().stream().map(TypeVariable::toString).collect(joining("" + SEMICOLON + NEW_LINE)));
+        query.append(rules().stream().map(Rule::toString).collect(joining("" + SEMICOLON + NEW_LINE)));
```
haikalpribadi, line 81: You've changed the order in which the user wrote the types and rules. If you just save the original list, and stream through that list in order, with an `if` switch to print different types, that would preserve the order, right?

java/query/GraqlDefinable.java
```diff
@@ -76,7 +88,7 @@ public final boolean equals(final Object o) {
         if (this == o) return true;
         if (o == null || getClass() != o.getClass()) return false;
         final GraqlDefinable that = (GraqlDefinable) o;
-        return this.keyword.equals(that.keyword) && this.variables.equals(that.variables);
+        return this.keyword.equals(that.keyword) && this.typeVariables.equals(that.typeVariables) && this.rules.equals(that.rules);
```
haikalpribadi, line 91: Do we not care about the order they were written now in determining their identity? Previously we did, right? If we save the initial list, and simply compare that, then we keep that behaviour.

haikalpribadi (changes requested):

java/query/GraqlDefinable.java
flyingsilverfin: was debating those choices with myself too! ok

java/query/GraqlDefinable.java
flyingsilverfin: its not - we also need to split out the definables into `rule` and `variables` - the verification loop is used to verify variables and inner variables (within constraints) are labeled instead of named variables. I tried one loop but to do verification and recording in one loop leads to awkward checks, or incorrectly recording labeled variables as top level type variables

flyingsilverfin: ive pushed a change that might make this clearer

java/query/GraqlDefinable.java
flyingsilverfin: yeah ok can keep the original list

java/query/GraqlDefinable.java
flyingsilverfin: can just use `Definable` list and use `toString` on `Definable`!

grammar/Graql.g4
haikalpribadi: okay

java/pattern/schema/Rule.java
```diff
@@ -0,0 +1,112 @@
...
+    @Override
+    public boolean equals(final Object o) {
+        if (o == this) {
+            return true;
+        }
+        if (o instanceof Rule) {
+            final Rule that = (Rule) o;
+            return this.label.equals(that.label) && this.when.equals(that.when) && this.then.equals(that.then);
+        }
+        return false;
```
haikalpribadi, line 110: Let's just use the common `equals()` method structure. You missed to check if the `Object o` is `null` here.

haikalpribadi (changes requested):

java/pattern/schema/Rule.java
haikalpribadi:
```
if (this == o) return true;
if (o == null || getClass() != o.getClass()) return false;
final Rule that = (Rule) o;
return (this.label.equals(that.label) && 
            this.when.equals(that.when) && 
            this.then.equals(that.then));
```

haikalpribadi: That's what it should be @flyingsilverfin

haikalpribadi: Why is this marked as resolved, @flyingsilverfin ? Is the change perhaps not pushed yet? Better to resolve comments after they've been pushed.

haikalpribadi: I'm sorry it's GitHub not refreshing the page properly even after I hit refresh 5 times! This has been resolved I can see now.

haikalpribadi (approved):

PR #164, opened 2020-10-14
Fix disjunction normalisation - missing sub-pattern cases

flyingsilverfin:
Fix normalisation in Graql patterns - every match or insert pattern should be normalisable.

* Add missing case of `pattern.isVariable()` and `pattern.isNegation()` when normalising disjunctions, which led to casting errors.
* Graql BDD steps always try to normalise queries when possible.

### DIFF
--- a/java/pattern/Disjunction.java
+++ b/java/pattern/Disjunction.java
@@ -51,8 +51,9 @@ public class Disjunction<T extends Pattern> implements Pattern {
     public Disjunction<Conjunction<Conjunctable>> normalise() {
         if (normalised == null) {
             final List<Conjunction<Conjunctable>> conjunctions = patterns.stream().flatMap(p -> {
-                if (p.isConjunction()) return Stream.of(new Conjunction<>(list(p.asConjunctable())));
+                if (p.isVariable()) return Stream.of(new Conjunction<>(list(p.asConjunctable())));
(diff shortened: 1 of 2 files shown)


java/pattern/Disjunction.java
```diff
@@ -51,8 +51,9 @@ public Disjunction(final List<T> patterns) {
     public Disjunction<Conjunction<Conjunctable>> normalise() {
         if (normalised == null) {
             final List<Conjunction<Conjunctable>> conjunctions = patterns.stream().flatMap(p -> {
-                if (p.isConjunction()) return Stream.of(new Conjunction<>(list(p.asConjunctable())));
+                if (p.isVariable()) return Stream.of(new Conjunction<>(list(p.asConjunctable())));
                 else if (p.isConjunction()) return p.asConjunction().normalise().patterns().stream();
+                else if (p.isNegation()) return Stream.of(new Conjunction<>(list(p.asNegation().normalise().asConjunctable())));
```
flyingsilverfin, line 56: not sure about this one...

flyingsilverfin: this seems to imply that a Conjunction can contain a normalised negation, which in general is also in Disjunction<Conjunction<...>> form?

haikalpribadi (approved):
👍

haikalpribadi (approved):

PR #169, opened 2020-10-23
Implement Rule Validation Checks

maxbaxt:

Implements the following rule validation:

- No nested negation in the **when** of a rule
- No disjunction present in the **when** of a rule
- rule **then** must contain contain either 1 has constraint, or an isa and relation constraint.

- two helper methods in *Rule.java* to help find negations and disjunctions
- all patterns now have a *patterns* methods to return all inner patters. This is useful for the helper methods mentioned above
- cleaned up error messages in order to make the code clean, and also improve the error messages given to the user


haikalpribadi:
This PR does not run on Grabl so we don't know if your changes pass an of the tests, @maxbaxt . You need to add your fork repos onto Grabl.

java/common/exception/ErrorMessage.java
```diff
@@ -53,20 +53,28 @@
...
+    public static final ErrorMessage INVALID_RULE_THEN_ZERO_CONSTRAINTS =
+            new ErrorMessage(20, "Rule '%s' 'then' must contain at least one constraint: '%s'.");
     public static final ErrorMessage INVALID_RULE_THEN_ONE_CONSTRAINT =
-            new ErrorMessage(18, "Rule '%s' 'then' with one constraint must be a new (or extended) relation, attribute ownership, or 'isa' downcast: '%s'.");
+            new ErrorMessage(21, "Rule '%s' 'then' with one constraint must be an attribute ownership: '%s'.");
     public static final ErrorMessage INVALID_RULE_THEN_TWO_CONSTRAINTS =
-            new ErrorMessage(19, "Rule '%s' 'then' with two constraints must must be a relation and an isa constraint: '%s'.");
+            new ErrorMessage(22, "Rule '%s' 'then' with two constraints must must be a relation and an isa constraint: '%s'.");
+    public static final ErrorMessage INVALID_RULE_THEN_THREE_OR_MORE_CONSTRAINT =
+            new ErrorMessage(23, "Rule '%s' 'then' must cannot contain more than three constraint: '%s'.");
```
haikalpribadi, line 67: Should we really split this error message to 3 different error message? How do you know that the user intended to put only 1 or 2 specifically? Telling them that they can only insert attribute ownership for 1 constraint, and relation an `isa` for 2 constraints, assume that those are the user's intentions. I think we should keep 1 generic error message for `then` violations.

haikalpribadi (changes requested):

java/common/exception/ErrorMessage.java
maxbaxt: Good point. I was thinking in terms of the code and not the way the user will see it.

haikalpribadi (approved):

PR #170, opened 2020-10-28
Correct Parsing and Validation for Attributes constraints

maxbaxt:

An attribute now has three options:

1. Attribute ownership including attribute type and a variable value (eg $x has age $a)
2. Attribute ownership including attribute type and a concrete value (ag $x has age 18)
3. Attribute ownership (eg $x has $a)

Number 1 is syntactic sugar for two constraints ($x has $a; $a 18 isa age;) and so cannot be used as a valid rule `then`.

Parser and Grammar updated to implement new restrictions. Rule validation changed to check correct --has attribute-- form used in rule heads.


java/common/exception/ErrorMessage.java
```diff
@@ -58,17 +58,20 @@
     public static final ErrorMessage INVALID_RULE_WHEN_CONTAINS_DISJUNCTION =
             new ErrorMessage(19, "Rule '%s' 'when' contains a disjunction.");
     public static final ErrorMessage INVALID_RULE_THEN =
-            new ErrorMessage(20, "Rule '%s' 'then' must strictly be either an attribute ownership, or a relation: '%s'.");
+            new ErrorMessage(20, "Rule '%s' 'then' must strictly be a single of either an attribute ownership, or a relation");
+    public static final ErrorMessage INVALID_RULE_THEN_HAS =
+            new ErrorMessage(21, "Rule '%s' 'then' is trying to assign both an attribute type and a variable attribute value: '%s'." +
```
flyingsilverfin, line 63: can we try our best to keep this to one line?

java/parser/Parser.java
```diff
@@ -741,13 +741,11 @@ public TypeVariable visitVariable_type(final GraqlParser.Variable_typeContext ct
...
-            return new ThingConstraint.Has(ctx.label().getText(), getVar(ctx.VAR_()));
-        } else if (ctx.value() != null) {
-            return new ThingConstraint.Has(ctx.label().getText(), visitValue(ctx.value()));
-        } else {
-            throw new IllegalArgumentException("Unrecognised MATCH HAS statement: " + ctx.getText());
-        }
+        if (ctx.label() != null) {
+            if (ctx.VAR_() != null) return new ThingConstraint.Has(ctx.label().getText(), getVar(ctx.VAR_()));
+            if (ctx.value() != null) return new ThingConstraint.Has(ctx.label().getText(), visitValue(ctx.value()));
+        } else if (ctx.VAR_() != null) return new ThingConstraint.Has(getVar(ctx.VAR_()));
```
flyingsilverfin, line 747: looks nice! :D

java/pattern/constraint/ThingConstraint.java
```diff
@@ -387,16 +388,19 @@ public Has(final String type, final UnboundVariable var) {
...
             this.hash = Objects.hash(Has.class, this.type, this.attribute);
         }
 
-        public ThingVariable<?> attribute() {
-            return attribute;
-        }
+
+        public ThingVariable<?> attribute() { return attribute; }
+
+        public TypeVariable type() { return type; }
```
flyingsilverfin, line 403: lets add @Nullable for the return type here

java/pattern/schema/Rule.java
```diff
@@ -118,6 +119,7 @@ private static void validateWhen(String label, Conjunction<? extends Pattern> wh
         return pattern.patterns().stream().flatMap(Rule::findDisjunctions);
     }
 
+    //TODO check if the change in INVALID_RULE_THEN_ONE_CONSTRAINT validation requires reworking of logic here or elsewhere
```
flyingsilverfin, line 122: do we still need this comment?

java/pattern/schema/Rule.java
```diff
@@ -127,6 +129,11 @@ private static void validateThen(String label, @Nullable Conjunction<? extends P
             throw GraqlException.of(INVALID_RULE_THEN.message(label, then));
         }
 
+        // rule 'has' cannot assign both an attribute type and a variable
+        if (numConstraints == 1 && then.has().size()==1 && then.has().get(0).type().isNamed() && then.has().get(0).attribute().isVariable()) {
```
flyingsilverfin, line 133: lets be consistent with spacing between the `==`. In one case you have spaces on either side, in the other you don't. It's a minor habit, bit if we're through and clean in execution we make our lives easier everywhere else :)

java/pattern/schema/Rule.java
```diff
@@ -127,6 +129,11 @@ private static void validateThen(String label, @Nullable Conjunction<? extends P
...
+            throw GraqlException.of(INVALID_RULE_THEN_HAS.message(label, then));
```
flyingsilverfin, line 134: so at this point we can already assume that if there's a `has`, then there is exactly 1 constraint right? so the first check is redundant

does this pass a test you can write that should fail? i think there maybe a missing `!`, and im not sure off the top of my head if `isNamed()` is correct, or not. Now that I think about it, the `isNamed()` is for when the user writes `$abc`, and  you can check if the variable is hidden (ie not named by user) and `variable.label()` exists to check for user-written label for a variable... tho might be a better way to do that

flyingsilverfin (changes requested):

java/common/exception/ErrorMessage.java
maxbaxt: I can but it makes the line super long.

java/parser/Parser.java
maxbaxt: th-thanks

java/pattern/schema/Rule.java
maxbaxt: Should have just done .type!=null.

I think attribute.isVariable is still true though?

java/pattern/constraint/ThingConstraint.java
maxbaxt: Done

java/pattern/schema/Rule.java
maxbaxt: deleted

java/pattern/schema/Rule.java
maxbaxt: Done

flyingsilverfin:
we still have failures?

java/common/exception/ErrorMessage.java
```diff
@@ -58,17 +58,19 @@
     public static final ErrorMessage INVALID_RULE_WHEN_CONTAINS_DISJUNCTION =
             new ErrorMessage(19, "Rule '%s' 'when' contains a disjunction.");
     public static final ErrorMessage INVALID_RULE_THEN =
-            new ErrorMessage(20, "Rule '%s' 'then' must strictly be either an attribute ownership, or a relation: '%s'.");
+            new ErrorMessage(20, "Rule '%s' 'then' must strictly be a single of either an attribute ownership, or a relation");
+    public static final ErrorMessage INVALID_RULE_THEN_HAS =
+            new ErrorMessage(21, "Rule '%s' 'then' is trying to assign both an attribute type and a variable attribute value: '%s'. Either make the attribute value concrete, or remove the attribute type.");
```
flyingsilverfin, line 63: so to make this shorter, we could write for instance:
"Rule '%s' cannot assign ownership of a variable attribute '%s' as well as a label '%s'.

I don't exactly like it but I think we can find more succinct language 👍 

who'dve thunk you'd be iterating on sentence structure in a logic-based programming job eeh!?

java/pattern/schema/Rule.java
flyingsilverfin: i think we resolved this on the phone right?

maxbaxt: Yes we did :)

flyingsilverfin (approved):
Don't forget to change back the git repository to mainline behaviour!

java/common/exception/ErrorMessage.java
```diff
@@ -62,17 +62,19 @@
     public static final ErrorMessage INVALID_RULE_WHEN_CONTAINS_DISJUNCTION =
             new ErrorMessage(21, "Rule '%s' 'when' contains a disjunction.");
     public static final ErrorMessage INVALID_RULE_THEN =
-            new ErrorMessage(22, "Rule '%s' 'then' must strictly be either an attribute ownership, or a relation: '%s'.");
+            new ErrorMessage(22, "Rule '%s' 'then' must strictly be a single of either an attribute ownership, or a relation");
+    public static final ErrorMessage INVALID_RULE_THEN_HAS =
+            new ErrorMessage(23, "Rule '%s' 'then' is trying to assign both an attribute type and a variable attribute value: '%s'. Either make the attribute value concrete, or remove the attribute type.");
```
flyingsilverfin, line 67: awh shiet sorry i forgot to say that you should also shorten this message like in the other PR so it isn't as verbose :) I'm sure you can reword it to fit in the character limit for one line!!

maxbaxt: shortened!

alexjpwalker (approved):

java/common/exception/ErrorMessage.java
```diff
@@ -62,17 +62,19 @@
     public static final ErrorMessage INVALID_RULE_WHEN_CONTAINS_DISJUNCTION =
             new ErrorMessage(21, "Rule '%s' 'when' contains a disjunction.");
     public static final ErrorMessage INVALID_RULE_THEN =
-            new ErrorMessage(22, "Rule '%s' 'then' must strictly be either an attribute ownership, or a relation: '%s'.");
+            new ErrorMessage(22, "Rule '%s' 'then' must strictly be a single of either an attribute ownership, or a relation");
```
haikalpribadi, line 65: The sentence contains a mistake here: "must strictly be a single ..." ? Single "something" ?

java/pattern/constraint/ThingConstraint.java
haikalpribadi: We don't do `@Nullable` for return type @flyingsilverfin , @maxbaxt . We only do You may notice from the rest of the code base that we only do `@Nullable` for method parameters, and if a return type of a getter could possibly be `null`, then we wrap it in a `Optional` -- which is what `Optional` is made for.

java/pattern/constraint/ThingConstraint.java
```diff
@@ -426,7 +432,7 @@ public boolean equals(final Object o) {
             if (this == o) return true;
             if (o == null || getClass() != o.getClass()) return false;
             final Has that = (Has) o;
-            return (this.type.equals(that.type) && this.attribute.equals(that.attribute));
+            return Objects.equals(this.type, that.type) && Objects.equals(this.attribute, that.attribute);
```
haikalpribadi, line 435: Why did we change `this.attribute.equals(that.attribute)` to using `Objects.equals()`? `Objects.equals()` is only used when the object you want to compare could possible be `null`. Which is why we should only use it for `this.type`.

java/pattern/schema/Rule.java
haikalpribadi: `then.has().get(0).type()` should be checked with `.isPresent()` if it were an `Optional`, which it should be.

haikalpribadi (changes requested):
Several revisions to go through, @maxbaxt. cc @flyingsilverfin

java/common/exception/ErrorMessage.java
maxbaxt: It means a single of either attribute ownership or a relation. In other words, either of those things, and exactly one instance of it. Reworded to make this clearer.

java/pattern/constraint/ThingConstraint.java
maxbaxt: Because we thought it looked neater to use the same method for this.attribute and this.type. Makes it easier to read. You are indeed correct we can remain with .equals for attribute.

maxbaxt: Happy to change it back to.equals though

java/common/exception/ErrorMessage.java
```diff
@@ -62,17 +62,19 @@
     public static final ErrorMessage INVALID_RULE_WHEN_CONTAINS_DISJUNCTION =
             new ErrorMessage(21, "Rule '%s' 'when' contains a disjunction.");
     public static final ErrorMessage INVALID_RULE_THEN =
-            new ErrorMessage(22, "Rule '%s' 'then' must strictly be either an attribute ownership, or a relation: '%s'.");
+            new ErrorMessage(22, "Rule '%s' 'then' must be exactly one attribute ownership, or exactly one relation: '%s'.");
+    public static final ErrorMessage INVALID_RULE_THEN_HAS =
+            new ErrorMessage(23, "Rule '%s' 'then' is trying to assign both an attribute type and a variable attribute value: '%s'.");
```
haikalpribadi, line 67: Putting the `then` clause that the user wrote at the right end of the message implies that that is what you're suggesting to be the right way, as opposed to pointing out that that is wrong. A better way is to have the content of the `then` clause to be mentioned right the start, such as: `"Rule '%s' 'then' ('%s') ..."`.

haikalpribadi (changes requested):
One more change on the error message.

java/common/exception/ErrorMessage.java
maxbaxt: Fixed those error messages.

haikalpribadi (approved):

PR #171, opened 2020-11-10
Fixed parsing of played-role overriding, and printing of overridden types

maxbaxt:

Fix a bug whereby Graql would not parse overriding of roles and attributes, in patterns such as `$x owns age as biometric;` , `$r relates father as parent;` etc

Parser was changed to correctly parse the overridden type. Three new tests were written in ParserTest to test overriding in relates, own and plays respectively. 

This exposed bugs in their respective toString method, which has now been fixed

### DIFF
--- a/java/parser/test/ParserTest.java
+++ b/java/parser/test/ParserTest.java
@@ -599,6 +599,66 @@ public class ParserTest {
     }
 
     @Test
+    public void whenParsingDefineQueryWithOwnsOverrides_ResultIsSameAsJavaGraql() {
+        final String query = "define\n" +
+                "triangle sub entity;\n" +
+                "triangle owns side-length;\n" +
+                "triangle-right-angled sub triangle;\n" +
+                "triangle-right-angled owns hypotenuse-length as side-length;";
+        final GraqlDefine parsed = Graql.parseQuery(query).asDefine();
+
+        final GraqlDefine expected = define(
+                type("triangle").sub("entity"),
+                type("triangle").owns("side-length"),
+                type("triangle-right-angled").sub("triangle"),
+                type("triangle-right-angled").owns("hypotenuse-length", "side-length")
+        );
+        assertQueryEquals(expected, parsed, query);
+    }
+
+    @Test
+    public void whenParsingDefineQueryWithRelatesOverrides_ResultIsSameAsJavaGraql() {
+        final String query = "define\n" +
+                "pokemon sub entity;\n" +
+                "evolves sub relation;\n" +
+                "evolves relates from, relates to;\n" +
+                "evolves-final sub evolves;\n" +
+                "evolves-final relates from-final as from;";
+        final GraqlDefine parsed = Graql.parseQuery(query).asDefine();
+
+        final GraqlDefine expected = define(
+                type("pokemon").sub("entity"),
+                type("evolves").sub("relation"),
+                type("evolves").relates("from").relates("to"),
+                type("evolves-final").sub("evolves"),
+                type("evolves-final").relates("from-final", "from")
+        );
+        assertQueryEquals(expected, parsed, query);
+    }
+
+    @Test
+    public void whenParsingDefineQueryWithPlaysOverrides_ResultIsSameAsJavaGraql() {
+        final String query = "define\n" +
+                "pokemon sub entity;\n" +
+                "evolves sub relation;\n" +
+                "evolves relates from, relates to;\n" +
+                "evolves-final sub evolves;\n" +
+                "evolves-final relates from-final as from;\n" +
+                "pokemon plays evolves-final:from-final as from;";
+        final GraqlDefine parsed = Graql.parseQuery(query).asDefine();
+
+        final GraqlDefine expected = define(
+                type("pokemon").sub("entity"),
+                type("evolves").sub("relation"),
+                type("evolves").relates("from").relates("to"),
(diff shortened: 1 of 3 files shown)


java/parser/test/ParserTest.java
```diff
@@ -598,6 +598,66 @@ public void whenParsingAsInMatch_ResultIsSameAsSub() {
         assertQueryEquals(expected, parsed, query);
     }
 
+    @Test
+    public void whenParsingDefineQueryWithOwnsOverrides_ResultIsSameAsJavaGraql() {
+        final String query = "define\n" +
+                "triangle sub entity;\n" +
+                "triangle owns side-length;\n" +
+                "triangle-right-angled sub triangle;\n" +
+                "triangle-right-angled owns hypotenuse-length as side-length;";
```
flyingsilverfin, line 607: nice example!

java/pattern/constraint/TypeConstraint.java
```diff
@@ -466,7 +466,17 @@ public boolean isOwns() {
 
         @Override
         public String toString() {
-            return OWNS.toString() + SPACE + attributeType + (isKey ? "" + SPACE + IS_KEY : "");
+            StringBuilder syntax = new StringBuilder();
+            syntax.append(OWNS).append(SPACE);
+            if (!attributeType.label().isPresent()) syntax.append(attributeType);
+            else syntax.append(attributeType.label().get().label());
+            if (isKey) syntax.append(SPACE).append(IS_KEY);
+            if (overriddenAttributeType!=null) {
```
flyingsilverfin, line 474: i think we prefer the more spaced out style with spaces on either side of the inequality, or whichever intelliJ's auto format performs

flyingsilverfin (approved):

java/pattern/constraint/TypeConstraint.java
maxbaxt: If statements duly socially distanced.

haikalpribadi: > i think we prefer the more spaced out style with spaces on either side of the inequality, or whichever intelliJ's auto format performs

Do we? :D I think we prefer to make code as compact as possible (without going over 120 character width limit)

java/pattern/constraint/TypeConstraint.java
```diff
@@ -466,7 +466,22 @@ public boolean isOwns() {
 
         @Override
         public String toString() {
-            return OWNS.toString() + SPACE + attributeType + (isKey ? "" + SPACE + IS_KEY : "");
+            StringBuilder syntax = new StringBuilder();
+            syntax.append(OWNS).append(SPACE);
+
+            if (!attributeType.label().isPresent()) syntax.append(attributeType);
+            else syntax.append(attributeType.label().get().label());
```
haikalpribadi, line 473: This logic is not needed given that `TypeVariable.toString()` already takes care of this. You can infer this from the fact that the previous code worked without this new logic.

java/pattern/constraint/TypeConstraint.java
```diff
@@ -466,7 +466,22 @@ public boolean isOwns() {
...
+
+            if (isKey) syntax.append(SPACE).append(IS_KEY);
+
+            if (overriddenAttributeType!=null) {
+                syntax.append(SPACE).append(AS).append(SPACE);
+
+                if (!overriddenAttributeType.label().isPresent()) syntax.append(overriddenAttributeType);
+                else syntax.append(overriddenAttributeType.label().get().label());
```
haikalpribadi, line 481: Same as above: this logic is not needed given that `TypeVariable.toString()` already takes care of this.

java/pattern/constraint/TypeConstraint.java
```diff
@@ -562,7 +577,24 @@ public boolean isPlays() {
...
+
+            syntax.append(PLAYS).append(SPACE);
+            if (relationType!=null) {
+                if (!relationType.label().isPresent()) syntax.append(relationType);
+                else syntax.append(relationType.label().get().label());
+                syntax.append(COLON);
+            }
+
+            if (!roleType.label().isPresent()) syntax.append(roleType);
+            else syntax.append(roleType.label().get().label());
```
haikalpribadi, line 590: This block is also entirely not needed. The old code covers this logic.

java/pattern/constraint/TypeConstraint.java
```diff
@@ -562,7 +577,24 @@ public boolean isPlays() {
...
+
+            if (overriddenRoleType != null) {
+                syntax.append(SPACE).append(AS).append(SPACE);
+                if (!overriddenRoleType.label().isPresent()) syntax.append(overriddenRoleType);
+                else syntax.append(overriddenRoleType.label().get().label());
```
haikalpribadi, line 595: Same as above: this logic is not needed given that `TypeVariable.toString()` already takes care of this.

haikalpribadi (changes requested):

java/pattern/constraint/TypeConstraint.java
maxbaxt: Done, reduced to the old style while keeping the fix.

java/pattern/constraint/TypeConstraint.java
maxbaxt: Fixed as above.

haikalpribadi (approved):

PR #190, opened 2021-01-19
Scope unscoped role types in Plays TypeConstraints

Phillammon:

When a role type is overridden within the Plays constraint, the overridden type is specified with no scope- we can infer the scope based on it being a supertype of the supplied role type, but at parse time this is not possible to infer without building the entire graph tree. In order to avert this, we supplied the root Relation type as a scope for the relation so it can be identified as a role type ahead of time, which leads to it being passed over for definition.

We added the root RelationType as the scope for overridden RoleTypes in the Plays TypeConstraint.
We changed the way that the Plays TypeConstraint reports its RoleTypes to match this inferred type.


java/pattern/constraint/TypeConstraint.java
```diff
@@ -565,8 +566,17 @@ public boolean isPlays() {
...
-            return PLAYS.toString() + SPACE + roleType +
-                    (overriddenRoleType != null ? "" + SPACE + AS + SPACE + overriddenRoleType : "");
+            String returnString = PLAYS.toString() + SPACE + roleType;
+            if (overriddenRoleType != null) {
+                String overriddenRoleTypeString;
+                if (overriddenRoleType.label().isPresent()) {
+                    overriddenRoleTypeString = overriddenRoleType.label().get().label();
+                } else {
+                    overriddenRoleTypeString = overriddenRoleType.reference().syntax();
+                }
```
flyingsilverfin, line 576: i wonder if we can compress this into a one liner with the `?` syntax and itll fit? probably not, but worth a try to see if it looks nicer :)

java/pattern/constraint/TypeConstraint.java
```diff
@@ -565,8 +566,17 @@ public boolean isPlays() {
...
+                returnString += "" + SPACE + AS + SPACE + overriddenRoleTypeString;
+            }
+            return returnString;
```
flyingsilverfin, line 579: so if we don't have an established name for the string we build up and return, i think we can improve `returnString` to be something more meaningful, such as `syntax`

flyingsilverfin (changes requested):
see comments inline

flyingsilverfin:
be careful with your titles - why is what you're doing specific to `define` queries? I believ its generic to any match or define  or undefine query actually, since you directly modified constraints. I think thats probably forgivable because you dont have a mental model yet of how Graql works. However, as you're writing your title and description is a good time to think and do one last check of your basic assumptions and understanding - maybe you could've questioned whether this was just for `Define`?.

As per the other PRs, add milestone, project, and labels plz

flyingsilverfin (approved):

haikalpribadi (approved):

haikalpribadi:
I've approved, but make sure to bump the dependency to graknlabs/behaviour repository first

PR #195, opened 2021-01-26
Mark Isa constraints derived via has constraints as derived

Phillammon:

In order to handle deletion properly, we need to be able to distinguish Isa constraints placed implicitly as part of has. For example, `$x has name $y` implies `$y isa name`, but if we are deleting `$x has name $y` we should not also delete $y's isa constraint. To this end, we added a flag on the Isa constraint indicating if it was derived.

Added derived boolean to Isa constraint, set only when creating an isa implicitly via the Has constraints.


flyingsilverfin (approved):

flyingsilverfin:
don't forget projects + labels + milestones!

haikalpribadi (approved):

PR #198, opened 2021-02-10
Create GraqlMatch.Modifier to group filters

flyingsilverfin:
To group together all the options available on a Match query, we create a `Modifier` object to hold `get` filter, and optional `limit`, `offset`, and `sort`

Create `GraqlMatch.Modifier` inner class to hold onto the modifiers available on a match query.

### DIFF
--- a/java/query/GraqlMatch.java
+++ b/java/query/GraqlMatch.java
@@ -64,7 +64,7 @@ import static java.util.stream.Stream.of;
 public class GraqlMatch extends GraqlQuery implements Aggregatable<GraqlMatch.Aggregate> {
 
     private final Conjunction<? extends Pattern> conjunction;
-    private final Modifier modifier;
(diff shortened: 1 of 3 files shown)


java/query/GraqlMatch.java
```diff
@@ -81,22 +80,91 @@
...
+
+        @Override
+        public int hashCode() {
+            return hash;
+        }
+    }
+
+    private List<UnboundVariable> retrievedVars() {
+        if (modifier.filter.isEmpty()) return variablesNamedUnbound;
+        else return modifier.filter;
```
flyingsilverfin, line 167: in the internal query validation, we used to first check filter vars if they exist, otherwise use the named vars. This is exposed as a new match-level method `retrievedVars()`

haikalpribadi (approved):

Issue #202, opened 2021-03-01
Type labels allow more special characters


flyingsilverfin:
## Problem to Solve
To enable more auto-generated and auto-imported data sources, we can allow an extended set of special characters inside a type label. The currently allowed set only includes `-` and `_`. We propose adding the following list of special characters for non-starting characters in a label.
* `+`
* `/`
* `[`, `]`
* `(`, `)`

## Current Workaround

Applications have to use a special encoding mapping like `__leftsq__` to swap in and out `[`.

## Proposed Solution
Add the proposed list to the allowed set of characters in a type label.

haikalpribadi:
Can you try submitting a PR for this, @flyingsilverfin? Let's see if it breaks anything in the grammar/parser.

haikalpribadi:
I shall close this issue too @flyingsilverfin

PR #204, opened 2021-03-05
Relax label syntax restrictions

flyingsilverfin:

To enable #202, we relax the grammar to allow special characters beyond `-` and `_` in type labels. However, we cannot add brackets of any kind `[]()` because it would clash with the remainder of the grammar, required to parse the language.

* add `+`, `/`, to the grammar for labels


.grabl/automation.yml
```diff
@@ -50,10 +50,10 @@ build:
         dependencies/maven/update.sh
         git diff --exit-code dependencies/maven/artifacts.snapshot
         bazel run @graknlabs_dependencies//tool/unuseddeps:unused-deps -- list
-    test-graql-java:
+    test-graql:
       image: graknlabs-ubuntu-20.04
       command: |
-        bazel test //java/... --test_output=errors
```
haikalpribadi, line 56: Why did we need to do this, @flyingsilverfin ?

haikalpribadi:
I really can't seem to get comfortable with adding `+` and `/` into the label syntax, @flyingsilverfin :/ - it just doesn't feel right ...

PR #212, opened 2021-06-04
Attribute variables are mandatory

flyingsilverfin:

We identify a bug that allows writing `match "john";` as a valid query and resolve it by mandating variables for attribute constraints.

* make attribute variables mandatory


flyingsilverfin:
Closing to implement discussed solution in `typeql-lang-java`

PR #230, opened 2022-09-21
Allow sorting per-variable

flyingsilverfin:

TypeQL can support sorting each variable in either `Ascending` or `Descending` order independently. Previously, TypeQL only supported ordering all the variables in the same order.

* Update grammar and parser to allow a sort order per-variable in the `sort` block
* Update language builders to allow a different sort order per variable
* Update tests to reflect the new capabilities

### DIFF
--- a/java/parser/test/ParserTest.java
+++ b/java/parser/test/ParserTest.java
@@ -414,7 +414,7 @@ public class ParserTest {
                 "$x plays starring:actor;\n" +
                 "sort $x asc;";
         TypeQLMatch parsed = TypeQL.parseQuery(query).asMatch();
-        TypeQLMatch expected = match(var("x").plays("starring", "actor")).sort("x", "asc");
+        TypeQLMatch expected = match(var("x").plays("starring", "actor")).sort(pair("x", "asc"));
 
         assertQueryEquals(expected, parsed, query);
     }
--- a/java/parser/test/ParserTest.java
+++ b/java/parser/test/ParserTest.java
@@ -34,15 +34,18 @@ import com.vaticle.typeql.lang.query.TypeQLMatch;
 import com.vaticle.typeql.lang.query.TypeQLQuery;
 import com.vaticle.typeql.lang.query.TypeQLUndefine;
 import com.vaticle.typeql.lang.query.TypeQLUpdate;
-import java.time.LocalDate;
-import java.time.LocalDateTime;
-import java.util.Collections;
-import java.util.List;
 import org.hamcrest.Matchers;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
+
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.util.Collections;
+import java.util.List;
+
 import static com.vaticle.typedb.common.collection.Collections.list;
+import static com.vaticle.typedb.common.collection.Collections.pair;
 import static com.vaticle.typeql.lang.TypeQL.and;
 import static com.vaticle.typeql.lang.TypeQL.define;
 import static com.vaticle.typeql.lang.TypeQL.gte;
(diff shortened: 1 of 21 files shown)


java/parser/test/ParserTest.java
```diff
@@ -449,11 +452,11 @@ public void testGetSortOffsetLimit() {
         final String query = "match\n" +
                 "$x isa movie,\n" +
                 "    has rating $r;\n" +
-                "sort $r desc; offset 10; limit 10;";
+                "sort $r desc, $x asc; offset 10; limit 10;";
         TypeQLMatch parsed = TypeQL.parseQuery(query).asMatch();
         TypeQLMatch expected = match(
                 var("x").isa("movie").has("rating", var("r"))
-        ).sort(list("r"), "desc").offset(10).limit(10);
+        ).sort(pair("r", "desc"), list(pair("$x", "asc"))).offset(10).limit(10);
```
flyingsilverfin, line 459: new example of sorting multiple variables differently

java/query/builder/Sortable.java
```diff
@@ -21,42 +21,48 @@
...
 import static com.vaticle.typeql.lang.common.TypeQLToken.Char.COMMA_SPACE;
 import static com.vaticle.typeql.lang.common.TypeQLToken.Char.SPACE;
-import static com.vaticle.typeql.lang.common.exception.ErrorMessage.INVALID_SORTING_ORDER;
-import static java.util.stream.Stream.concat;
-import static java.util.stream.Stream.of;
+import static com.vaticle.typeql.lang.common.exception.ErrorMessage.VARIABLE_NOT_SORTED;
 
 public interface Sortable<S, O, L> {
 
     default S sort(String var, String... vars) {
```
flyingsilverfin, line 44: the basic entry point is the same as before, and assumes `ASC` for every variable

java/query/builder/Sortable.java
```diff
@@ -21,42 +21,48 @@
...
-        TypeQLArg.Order o = TypeQLArg.Order.of(order);
-        if (o == null) throw TypeQLException.of(
-                INVALID_SORTING_ORDER.message(TypeQLArg.Order.ASC, TypeQLArg.Order.DESC)
-        );
-        return sort(vars.stream().map(UnboundVariable::named).collect(Collectors.toList()), o);
-    }
-
-    default S sort(List<UnboundVariable> vars) {
-        return sort(new Sorting(vars));
+    default S sort(Pair<String, String> varOrder, List<Pair<String, String>> varOrders) {
```
flyingsilverfin, line 55: the secondary "shorthand" entry point is using pairs of strings, eg `pair("x", "asc")` to indicate a variable's order.

However, this clashes with the canonical API (below), so we split it into the two methods above, the second of which takes a list of following variable orderings

java/query/builder/Sortable.java
```diff
@@ -67,43 +73,49 @@ default S sort(List<UnboundVariable> vars, TypeQLArg.Order order) {
 
     class Sorting {
 
-        private final List<UnboundVariable> vars;
-        private final TypeQLArg.Order order;
         private final int hash;
+        private final List<UnboundVariable> variables;
+        private final Map<UnboundVariable, TypeQLArg.Order> orders;
```
flyingsilverfin, line 78: once the list of pairs is received, we normalise it into a `List` of variables and a `Map` of orders

java/query/builder/Sortable.java
```diff
@@ -67,43 +73,49 @@ default S sort(List<UnboundVariable> vars, TypeQLArg.Order order) {
...
-            return vars;
+        public List<UnboundVariable> variables() {
+            return variables;
         }
 
-        public TypeQLArg.Order order() {
-            return order == null ? TypeQLArg.Order.ASC : order;
+        public TypeQLArg.Order getOrder(UnboundVariable var) {
+            if (variables.contains(var)) throw TypeQLException.of(VARIABLE_NOT_SORTED.message(var));
+            return orders.getOrDefault(var, TypeQLArg.Order.ASC);
```
flyingsilverfin, line 102: we retain `null` in the order map to be able to remember which variables used the default order. Only when retrieving the order do we default it to `ASC`,  but we don't store `ASC` in the map as it would distort the `toString()`

haikalpribadi (approved):

PR #233, opened 2022-10-03
Split delete_or_update in the grammar into separate parser rules

dmitrii-ubskii:

We split the `delete` and `update` queries into separate parser rules in the TypeQL grammar, resolving a TODO in the process.

Split `delete` and `update` structures on grammar level, and modified the libraries accordingly.

Drive-by: remove unused `label_array` and `labels`.

### DIFF
--- a/grammar/TypeQL.g4
+++ b/grammar/TypeQL.g4
@@ -32,18 +32,17 @@ eof_schema_rule       :   schema_rule      EOF ;
 // TYPEQL QUERY LANGUAGE ========================================================
 
 query                 :   query_define      |   query_undefine
-                      |   query_update      |   query_insert      |   query_delete
+                      |   query_insert      |   query_update      |   query_delete
                       |   query_match       |   query_match_aggregate
                       |   query_match_group |   query_match_group_agg           ;
 
 query_define          :   DEFINE      definables  ;
 query_undefine        :   UNDEFINE    definables  ;
 
-query_update          :   MATCH       patterns      DELETE  variable_things
-                                                    INSERT  variable_things     ;
 query_insert          :   MATCH       patterns      INSERT  variable_things
                       |                             INSERT  variable_things     ;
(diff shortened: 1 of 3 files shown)


grammar/TypeQL.g4
```diff
@@ -32,18 +32,18 @@ eof_schema_rule       :   schema_rule      EOF ;
...
-                      |   query_insert      |   query_delete_or_update
+                      |   query_update      |   query_insert      |   query_delete
                       |   query_match       |   query_match_aggregate
                       |   query_match_group |   query_match_group_agg           ;
 
 query_define          :   DEFINE      definables  ;
 query_undefine        :   UNDEFINE    definables  ;
 
+query_update          :   MATCH       patterns      DELETE  variable_things
+                                                    INSERT  variable_things     ;
```
dmitrii-ubskii, line 43: I considered also `query_delete` followed by `INSERT variable_things`, but this reads better in my opinion.

grammar/TypeQL.g4
```diff
@@ -179,8 +177,6 @@ type                  :   label                         | VAR_          ;
 label_any             :   label_scoped  | label         ;
 label_scoped          :   LABEL_SCOPED_ ;
 label                 :   LABEL_        | schema_native | type_native   | unreserved    ;
-labels                :   label         | label_array   ;
-label_array           :   '[' label ( ',' label )* ']'  ;
```
dmitrii-ubskii, line 183: Not used anywhere. May be returned later if needed.

grammar/TypeQL.g4
flyingsilverfin: I actually think the `query_delete` followed by `INSERT` is better so that we end up with 1 definition of `delete`.
Then i'd also put the `update` right after the `insert` and before the `delete` definition.
The same order should be reflected in the top `query : ...`

flyingsilverfin (approved):

PR #235, opened 2022-10-10
Feature complete match query handling in TypeQL Rust

dmitrii-ubskii:

We complete the implementation of the match query building and parsing by adding handlers for inequality and string predicates and logical operations (and, or, not).

- Features:
  - implement Concept variables along with the `is` constraint;
  - implement Disjunction and Negation, finish Conjunction;
  - handle non-equality (non-`=`) predicates;
- Refactoring:
  - use concrete structs (not wrapped in enum) everywhere;
  - unify `Display` trait implementations;
  - clean up string formatting;
  - replace hardcoded strings with string enumerations for keywords and operators;
  - provide top-level builder macros that can handle arbitrary number of arguments that may be of different types.

### DIFF
--- a/rust/common/string.rs
+++ b/rust/common/string.rs
@@ -28,6 +28,10 @@ pub(crate) fn unquote_string(quoted_string: String) -> String {
     String::from(&quoted_string[1..quoted_string.len() - 1])
 }
 
+pub(crate) fn indent(multiline_string: String) -> String {
+    format!("    {}", multiline_string.replace('\n', "\n    "))
+}
+
 pub(crate) fn escape_regex(regex: &str) -> String {
     regex.replace('/', r#"\/"#)
 }
--- a/rust/common/string.rs
+++ b/rust/common/string.rs
@@ -20,15 +20,15 @@
  *
  */
 
-pub(crate) fn quote_string(string: String) -> String {
+pub(crate) fn quote(string: &str) -> String {
     format!("\"{}\"", string)
 }
 
-pub(crate) fn unquote_string(quoted_string: String) -> String {
+pub(crate) fn unquote(quoted_string: &str) -> String {
     String::from(&quoted_string[1..quoted_string.len() - 1])
 }
 
-pub(crate) fn indent(multiline_string: String) -> String {
+pub(crate) fn indent(multiline_string: &str) -> String {
     format!("    {}", multiline_string.replace('\n', "\n    "))
 }
 
@@ -36,6 +36,6 @@ pub(crate) fn escape_regex(regex: &str) -> String {
     regex.replace('/', r#"\/"#)
 }
 
-pub(crate) fn unescape_regex(regex: String) -> String {
+pub(crate) fn unescape_regex(regex: &str) -> String {
     regex.replace(r#"\/"#, "/")
 }
--- a/rust/common/string.rs
+++ b/rust/common/string.rs
@@ -20,6 +20,18 @@
  *
  */
 
-pub mod error;
-pub mod string;
-pub mod token;
+pub(crate) fn quote_string(string: String) -> String {
+    format!("\"{}\"", string)
+}
+
+pub(crate) fn unquote_string(quoted_string: String) -> String {
+    String::from(&quoted_string[1..quoted_string.len() - 1])
+}
+
+pub(crate) fn escape_regex(regex: &str) -> String {
+    regex.replace('/', r#"\/"#)
+}
+
+pub(crate) fn unescape_regex(regex: String) -> String {
+    regex.replace(r#"\/"#, "/")
+}
--- a/rust/parser/mod.rs
+++ b/rust/parser/mod.rs
@@ -336,7 +336,7 @@ fn visit_variable_concept(ctx: Rc<Variable_conceptContext>) -> ParserResult<Conc
 fn visit_variable_type(ctx: Rc<Variable_typeContext>) -> ParserResult<TypeVariable> {
     let mut var_type = match visit_type_any(ctx.type_any().unwrap())? {
         Type::Variable(p) => p,
-        Type::Label(p) => UnboundVariable::hidden().type_(p)?.into_type(),
+        Type::Label(p) => UnboundVariable::hidden().type_(p)?,
     };
     for constraint in (0..).map_while(|i| ctx.type_constraint(i)) {
         if constraint.OWNS().is_some() {
@@ -361,7 +361,7 @@ fn visit_variable_type(ctx: Rc<Variable_typeContext>) -> ParserResult<TypeVariab
                 },
             );
         } else if constraint.REGEX().is_some() {
-            var_type = var_type.regex(get_regex(constraint.STRING_().unwrap()))?.into_type();
+            var_type = var_type.regex(get_regex(constraint.STRING_().unwrap()))?;
         } else if constraint.RELATES().is_some() {
             let _overridden: Option<()> = match constraint.AS() {
                 None => None,
@@ -380,7 +380,7 @@ fn visit_variable_type(ctx: Rc<Variable_typeContext>) -> ParserResult<TypeVariab
                 });
         } else if constraint.TYPE().is_some() {
             let scoped_label = visit_label_any(constraint.label_any().unwrap())?;
-            var_type = var_type.type_(scoped_label)?.into_type();
+            var_type = var_type.type_(scoped_label)?;
         } else {
             panic!("visit_variable_type: not implemented")
         }
@@ -411,7 +411,7 @@ fn visit_variable_thing_any(ctx: Rc<Variable_thing_anyContext>) -> ParserResult<
 fn visit_variable_thing(ctx: Rc<Variable_thingContext>) -> ParserResult<ThingVariable> {
     let mut var_thing = get_var(ctx.VAR_().unwrap()).into_thing();
     if let Some(iid) = ctx.IID_() {
-        var_thing = var_thing.iid(iid.get_text())?.into_thing();
+        var_thing = var_thing.iid(iid.get_text())?;
     }
     if let Some(isa) = ctx.ISA_() {
         var_thing = var_thing.constrain_isa(get_isa_constraint(isa, ctx.type_().unwrap())?)
--- a/rust/parser/mod.rs
+++ b/rust/parser/mod.rs
@@ -290,8 +290,14 @@ fn visit_patterns(ctx: Rc<PatternsContext>) -> ParserResult<Vec<Pattern>> {
 fn visit_pattern(ctx: Rc<PatternContext>) -> ParserResult<Pattern> {
     if let Some(var) = ctx.pattern_variable() {
         Ok(visit_pattern_variable(var)?.into_pattern())
-    } else {
+    } else if let Some(_disjunction) = ctx.pattern_disjunction() {
         todo!()
+    } else if let Some(_conjunction) = ctx.pattern_conjunction() {
+        todo!()
+    } else if let Some(negation) = ctx.pattern_negation() {
+        Ok(visit_pattern_negation(negation)?.into_pattern())
+    } else {
+        Err(ILLEGAL_GRAMMAR.format(&[&ctx.get_text()]))
     }
 }
 
@@ -303,8 +309,12 @@ fn visit_pattern_disjunction(_ctx: Rc<Pattern_disjunctionContext>) -> ParserResu
     todo!()
 }
 
-fn visit_pattern_negation(_ctx: Rc<Pattern_negationContext>) -> ParserResult<()> {
-    todo!()
+fn visit_pattern_negation(ctx: Rc<Pattern_negationContext>) -> ParserResult<Negation> {
+    let mut patterns = visit_patterns(ctx.patterns().unwrap())?;
+    Ok(match patterns.len() {
+        1 => Negation::from(patterns.pop().unwrap()),
+        _ => Negation::from(Conjunction::from(patterns).into_pattern()),
+    })
 }
 
 fn visit_pattern_variable(ctx: Rc<Pattern_variableContext>) -> ParserResult<Variable> {
@@ -319,8 +329,8 @@ fn visit_pattern_variable(ctx: Rc<Pattern_variableContext>) -> ParserResult<Vari
     }
 }
 
-fn visit_variable_concept(_ctx: Rc<Variable_conceptContext>) -> ParserResult<TypeVariable> {
-    todo!()
+fn visit_variable_concept(ctx: Rc<Variable_conceptContext>) -> ParserResult<ConceptVariable> {
+    get_var(ctx.VAR_(0).unwrap()).is(get_var(ctx.VAR_(1).unwrap()))
 }
 
 fn visit_variable_type(ctx: Rc<Variable_typeContext>) -> ParserResult<TypeVariable> {
--- a/rust/parser/mod.rs
+++ b/rust/parser/mod.rs
@@ -506,6 +506,14 @@ fn visit_predicate(ctx: Rc<PredicateContext>) -> ParserResult<ValueConstraint> {
                 panic!("Unexpected predicate value: `{}`", predicate_value.get_text())
             }
         }))
+    } else if let Some(substring) = ctx.predicate_substring() {
+        Ok(ValueConstraint::new(Predicate::from(substring.get_text()), {
+            if let Some(_) = substring.LIKE() {
+                Value::from(get_regex(ctx.STRING_().unwrap()))
+            } else {
+                Value::from(get_string(ctx.STRING_().unwrap()))
+            }
+        }))
     } else {
         todo!()
     }
--- a/rust/parser/mod.rs
+++ b/rust/parser/mod.rs
@@ -33,12 +33,11 @@ use chrono::{NaiveDate, NaiveDateTime, Timelike};
 use std::rc::Rc;
 
 use crate::common::error::{ErrorMessage, ILLEGAL_GRAMMAR};
-use typeql_grammar::typeqlrustparser::*;
 use crate::common::token::Predicate;
+use typeql_grammar::typeqlrustparser::*;
 
 use crate::pattern::*;
 use crate::query::*;
-use crate::typeql_match;
 
 #[derive(Debug)]
 pub struct Definable;
@@ -206,7 +205,8 @@ fn visit_query_update(_ctx: Rc<Query_updateContext>) -> ParserResult<()> {
 }
 
 fn visit_query_match(ctx: Rc<Query_matchContext>) -> ParserResult<TypeQLMatch> {
-    let mut match_query = typeql_match(visit_patterns(ctx.patterns().unwrap())?)?;
+    let mut match_query =
+        TypeQLMatch::new(Conjunction::from(visit_patterns(ctx.patterns().unwrap())?));
     if let Some(modifiers) = ctx.modifiers() {
         if let Some(filter) = modifiers.filter() {
             match_query = match_query.filter(visit_filter(filter)?);
--- a/rust/parser/mod.rs
+++ b/rust/parser/mod.rs
@@ -53,11 +53,11 @@ enum Type {
 }
 
 fn get_string(string: Rc<TerminalNode>) -> String {
-    unquote_string(string.get_text())
+    unquote(&string.get_text())
 }
 
 fn get_regex(string: Rc<TerminalNode>) -> String {
-    unescape_regex(unquote_string(string.get_text()))
+    unescape_regex(&unquote(&string.get_text()))
 }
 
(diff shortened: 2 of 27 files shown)


grabl:
### PR Review Checklist

Do not edit the content of this comment. The PR reviewer should simply update this comment by ticking each review item below, as they get completed.

----
**Code**

- [ ] Packages, classes, and methods have a single domain of responsibility.
- [ ] Packages, classes, and methods are grouped into cohesive and consistent domain model.
- [ ] The code is canonical and the minimum required to achieve the goal.
- [ ] Modules, libraries, and APIs are easy to use, robust (foolproof and not errorprone), and tested.
- [ ] Logic and naming has clear narrative that communicates the accurate intent and responsibility of each module (e.g. method, class, etc.).
- [ ] The code is algorithmically efficient and scalable for the whole application.

**Architecture**

- [ ] Any required refactoring is completed, and the architecture does not introduce technical debt incidentally.
- [ ] Any required build and release automations are updated and/or implemented.
- [ ] Any new components follows a consistent style with respect to the pre-existing codebase.
- [ ] The architecture intuitively reflects the application domain, and is easy to understand.
- [ ] The architecture has a well-defined hierarchy of encapsulated components.
- [ ] The architecture is extensible and scalable.

rust/parser/mod.rs
```diff
@@ -250,11 +248,7 @@ fn visit_sort(ctx: Rc<SortContext>) -> Sorting {
 fn visit_var_order(ctx: Rc<Var_orderContext>) -> OrderedVariable {
     OrderedVariable {
         var: get_var(ctx.VAR_().unwrap()),
-        order: if let Some(order) = ctx.ORDER_() {
-            Some(order.get_text())
-        } else {
-            None
-        },
+        order: ctx.ORDER_().map(|order| order.get_text()),
```
dmitrii-ubskii, line 251: Same thing, but idiomatic Rust.

rust/parser/mod.rs
```diff
@@ -51,18 +52,10 @@ enum Type {
...
-}
-
 fn get_string(string: Rc<TerminalNode>) -> String {
     unquote_string(string.get_text())
 }
 
-fn unescape_regex(regex: String) -> String {
-    regex.replace(r#"\\/"#, "/")
-}
-
```
dmitrii-ubskii, line 65: Moved to common/string

rust/parser/mod.rs
```diff
@@ -322,14 +337,14 @@ fn visit_pattern_variable(ctx: Rc<Pattern_variableContext>) -> ParserResult<Vari
...
-    todo!()
+fn visit_variable_concept(ctx: Rc<Variable_conceptContext>) -> ParserResult<ConceptVariable> {
+    get_var(ctx.VAR_(0).unwrap()).is(get_var(ctx.VAR_(1).unwrap()))
 }
 
 fn visit_variable_type(ctx: Rc<Variable_typeContext>) -> ParserResult<TypeVariable> {
     let mut var_type = match visit_type_any(ctx.type_any().unwrap())? {
         Type::Variable(p) => p,
-        Type::Label(p) => UnboundVariable::hidden().type_(p)?.into_type(),
+        Type::Label(p) => UnboundVariable::hidden().type_(p)?,
```
dmitrii-ubskii, line 347: The tightened type system doesn't require wrapping into and unwrapping from the enums. Most `into_X()` invocations are made redundant as a result.

rust/parser/syntax_error.rs
```diff
@@ -20,7 +20,7 @@
  *
  */
 
-use std::fmt::{Display, Formatter};
+use std::fmt;
```
dmitrii-ubskii, line 23: The usage of `std::fmt::Display` is unified across the repo in line with major crates (tokio, chrono).

rust/parser/test/mod.rs
```diff
@@ -21,8 +21,9 @@
  */
```
dmitrii-ubskii, line 22: Most changes here are replacing the top-level builders with the corresponding macros and reformatting of uncommented tests.

rust/pattern/constraint/thing_constraint.rs
```diff
@@ -64,9 +67,9 @@ impl TryFrom<String> for IIDConstraint {
     }
 }
 
-impl Display for IIDConstraint {
+impl fmt::Display for IIDConstraint {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "iid {}", self.iid)
+        write!(f, "{} {}", Iid, self.iid)
```
dmitrii-ubskii, line 72: Many hardcoded keywords are replaced with elements of string enums from common/token.

rust/pattern/constraint/thing_constraint.rs
```diff
@@ -166,67 +169,19 @@ impl ValueConstraint {
...
 }
 
-impl Display for ValueConstraint {
+impl fmt::Display for ValueConstraint {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        if self.predicate == Predicate::Eq && !matches!(self.value, Value::Variable(_)) {
+        if self.predicate == Predicate::Like {
+            assert!(matches!(self.value, Value::String(_)));
+            write!(f, "{} {}", self.predicate, escape_regex(&self.value.to_string()))
+        } else if self.predicate == Predicate::Eq && !matches!(self.value, Value::Variable(_)) {
```
dmitrii-ubskii, line 177: `like` is a special case.

rust/pattern/constraint/thing_constraint.rs
```diff
@@ -166,67 +169,19 @@ impl ValueConstraint {
...
-    Eq,
-    Neq,
-    Gt,
-    Gte,
-    Lt,
-    Lte,
-    // substring
-    Contains,
-    Like,
-}
```
dmitrii-ubskii, line 191: Moved to common/token

rust/pattern/constraint/thing_constraint.rs
```diff
@@ -340,11 +297,11 @@ impl From<RolePlayerConstraint> for RelationConstraint {
...
 
-impl Display for RelationConstraint {
+impl fmt::Display for RelationConstraint {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.write_char('(')?;
-        write_joined!(f, self.role_players, ", ")?;
-        f.write_char(')')
+        f.write_str("(")?;
+        write_joined!(f, ", ", self.role_players)?;
+        f.write_str(")")
```
dmitrii-ubskii, line 298: `write_char` requires importing a trait; `write_str` has no such prerequisites. Cleaner this way, and seems to be in line with the way major crates do `Display`.

rust/pattern/constraint/type_constraint.rs
```diff
@@ -223,10 +216,10 @@ impl From<bool> for IsKey {
 
 pub const KEY: IsKey = IsKey::Yes;
 
-impl Display for IsKey {
-    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+impl fmt::Display for IsKey {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         if *self == IsKey::Yes {
-            f.write_str(" @key")?;
+            write!(f, " {}", IsKey)?;
```
dmitrii-ubskii, line 222: May be a little confusing: this is `token::Constraint::IsKey`, which shares its name with `enum IsKey` from this file. It's unambiguous from the compiler's point of view.

rust/pattern/constraint/type_constraint.rs
```diff
@@ -311,12 +301,8 @@ impl From<String> for RegexConstraint {
     }
 }
 
-fn escape_regex(regex: &str) -> String {
-    regex.replace('/', r#"\\/"#)
-}
```
dmitrii-ubskii, line 316: Moved to common/string

rust/pattern/variable/builder/thing_variable_builder.rs
```diff
@@ -36,52 +37,84 @@ pub trait ThingVariableBuilder {
...
-    fn iid<T: TryInto<IIDConstraint>>(self, iid: T) -> Result<Variable, ErrorMessage>
+    fn iid<T: TryInto<IIDConstraint>>(self, iid: T) -> Result<ThingVariable, ErrorMessage>
     where
         ErrorMessage: From<<T as TryInto<IIDConstraint>>::Error>;
 
-    fn isa(self, isa: impl Into<IsaConstraint>) -> Result<Variable, ErrorMessage>;
+    fn isa(self, isa: impl Into<IsaConstraint>) -> Result<ThingVariable, ErrorMessage>;
 
-    fn eq(self, value: impl Into<Value>) -> Result<Variable, ErrorMessage>;
+    fn eq(self, value: impl Into<Value>) -> Result<ThingVariable, ErrorMessage>;
```
dmitrii-ubskii, line 50: Always return a concrete struct instance in the builder functions, so that we can leverage static analysis and compile-time type checking and avoid dynamic type checks (e.g. `Variable` checking if it is a `Thing` or `Unbound` on an `isa()` call.

rust/pattern/variable/thing_variable.rs
```diff
@@ -89,41 +109,15 @@ impl RelationConstrainable for ThingVariable {
...
-                f.write_char(' ')?;
-            }
-            write!(f, "{}", relation)?;
-        }
-
-        if let Some(iid) = &self.iid {
-            write!(f, " {}", iid)?;
+        if self.is_thing_constrained() {
+            f.write_str(" ")?;
+            write_joined!(f, ",\n    ", self.isa, self.iid, self.has)?;
```
dmitrii-ubskii, line 118: More powerful `write_joined!` now allows us to string together sequences of different types. (Remember that `Option` is treated as a sequence of zero or one elements)

rust/util/mod.rs
```diff
@@ -34,11 +34,21 @@ macro_rules! enum_getter {
 
 #[macro_export]
 macro_rules! write_joined {
-    ($f:ident, $iterable:expr, $joiner:tt) => {{
+    ($f:ident, $joiner:expr, $($iterable:expr),* $(,)*) => {{
```
alexjpwalker, line 37: We could use an explanation of this new implementation too - I believe you mentioned already it allows us to, once again, pass in a varargs of heterogeneous types

alexjpwalker (changes requested):

rust/pattern/variable/builder/thing_variable_builder.rs
```diff
@@ -36,52 +37,84 @@ pub trait ThingVariableBuilder {
...
+    }
+
+    fn lte(self, value: impl Into<Value>) -> Result<ThingVariable, ErrorMessage> {
+        Ok(self.constrain_value(ValueConstraint::new(Predicate::Lte, value.into())))
     }
 
-    fn isa(self, isa: impl Into<IsaConstraint>) -> Result<Variable, ErrorMessage> {
-        Ok(self.constrain_isa(isa.into()).into_variable())
+    fn contains(self, value: impl Into<String>) -> Result<ThingVariable, ErrorMessage> {
+        Ok(self.constrain_value(ValueConstraint::new(Predicate::Contains, value.into().into())))
```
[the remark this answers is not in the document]
dmitrii-ubskii: The first `into()` comes from `Into<String>`, the second -- from `Into<Value>`

rust/common/string.rs
```diff
@@ -0,0 +1,41 @@
...
+
+pub(crate) fn unquote_string(quoted_string: String) -> String {
+    String::from(&quoted_string[1..quoted_string.len() - 1])
+}
+
+pub(crate) fn indent(multiline_string: String) -> String {
+    format!("    {}", multiline_string.replace('\n', "\n    "))
+}
+
+pub(crate) fn escape_regex(regex: &str) -> String {
```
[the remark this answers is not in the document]
dmitrii-ubskii: Good catch, everything in this module should accept `&str`, as a matter of fact.

rust/parser/mod.rs
dmitrii-ubskii: Do you mean things like `UnboundVariable::into_thing()` or `SomeConstraint::from::<UnboundVariable>()`? Both are needed, because the meaning behind a variable being unbound is that we don't yet know what kind of variable it is, and its context will help determine whether it's a thing, a type, or a concept.

Methods like `into_variable()` and `into_pattern()` are still needed for actual type erasure so that we can store all the different kinds of patterns in a homogenous collection, e.g. `Conjunction` contains a `Vec<Pattern>`.

rust/parser/syntax_error.rs
dmitrii-ubskii: I think that's just to reduce the number of imports.

rust/pattern/constraint/concept_constraint.rs
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+use crate::common::token::Constraint::Is;
+use crate::pattern::*;
+use crate::var;
+use std::fmt;
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct IsConstraint {
+    variable: Box<ConceptVariable>,
```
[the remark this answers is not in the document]
dmitrii-ubskii: Precisely that.

rust/pattern/negation.rs
```diff
@@ -0,0 +1,68 @@
...
+ *
+ */
+
+use crate::common::token::Operator::Not;
+use crate::{ErrorMessage, Pattern, Variable};
+use core::fmt;
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct Negation {
+    pub pattern: Box<Pattern>,
```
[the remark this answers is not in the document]
dmitrii-ubskii: `Box` is not inherently a bad thing, I don't think we need to necessarily avoid it. It's not any worse than the `Vec<_>` field in `Conjunction`.

rust/pattern/variable/builder/thing_variable_builder.rs
```diff
@@ -36,52 +37,84 @@ pub trait ThingVariableBuilder {
...
+    fn iid<T: TryInto<IIDConstraint>>(self, iid: T) -> Result<ThingVariable, ErrorMessage>
     where
         ErrorMessage: From<<T as TryInto<IIDConstraint>>::Error>,
     {
-        Ok(self.constrain_iid(iid.try_into()?).into_variable())
+        Ok(self.constrain_iid(iid.try_into()?))
+    }
+
+    fn isa(self, isa: impl Into<IsaConstraint>) -> Result<ThingVariable, ErrorMessage> {
+        Ok(self.constrain_isa(isa.into()))
```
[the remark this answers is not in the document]
dmitrii-ubskii: That was because a macro expects `into_pattern`. It should really be a trait or, more likely, an `impl From<T> for Pattern`.

It should be clear here what type we convert to. `isa` is `impl Into<IsaConstraint>`, so `isa.into()` therefore returns `IsaConstraint`. Bare `into()` calls appear to be idiomatic Rust.

rust/query/typeql_match.rs
```diff
@@ -82,58 +83,53 @@ impl Display for TypeQLMatch {
 
 #[derive(Debug, Eq, PartialEq, Default)]
 pub struct Modifiers {
-    pub filter: Vec<UnboundVariable>,
+    pub filter: Option<Filter>,
     pub sorting: Option<Sorting>,
-    pub limit: Option<usize>,
-    pub offset: Option<usize>,
+    pub limit: Option<Limit>,
+    pub offset: Option<Offset>,
```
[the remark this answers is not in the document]
dmitrii-ubskii: Precisely that.

rust/typeql_lang.rs
```diff
@@ -51,25 +49,58 @@ pub fn parse_query(typeql_query: &str) -> Result<Query, String> {
...
+        let patterns = [$($pattern.map(|p| p.into_pattern())),*].into_iter().collect::<Result<Vec<_>, ErrorMessage>>();
+        match patterns {
+            Ok(patterns) => Ok(TypeQLMatch::new(Conjunction::from(patterns))),
+            Err(err) => Err(err),
+        }
+    }}
+}
+
+#[macro_export]
+macro_rules! and {
```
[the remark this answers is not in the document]
dmitrii-ubskii: Yes, and as you may have noticed, they do exactly the same as `typeql_match!`. I tried to factor the common bits out, but it ended up even less readable, so WET code it is. At least for now.

rust/util/mod.rs
dmitrii-ubskii: That's pretty much that. The macro unrolls into consecutive code blocks roughly equivalent to the previous version, except we keep track of whether the current iterator is the first non-empty one, just so we don't start the output with the joiner.

One other change is that the joiner is now allowed to be an expression, which will be handy when we'll have computed joiners.

rust/parser/syntax_error.rs
alexjpwalker: Let's go back to importing `Display` and `Formatter` directly. At the cost of one extra import, we get terser code that we use _very_ regularly.

We frequently take inspiration from other open-source code, review their standards, and then see if these standards are the best possible according to our values. In this case, I believe `Display` and `Formatter` are minimal and non-ambiguous, which is our benchmark for naming.

rust/pattern/constraint/type_constraint.rs
```diff
@@ -200,33 +193,33 @@ impl From<TypeVariable> for PlaysConstraint {
...
-        write!(f, "plays {}", self.role_type)
+impl fmt::Display for PlaysConstraint {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{} {}", Plays, self.role_type)
     }
 }
 
 #[derive(Debug, Clone, Copy, Eq, PartialEq)]
-pub enum IsKey {
+pub enum IsKeyAttribute {
```
alexjpwalker, line 203: Actually, I think `IsKeyConstraint` is a better name - we never call `@key` a "key attribute", we call it a key constraint

rust/pattern/variable/builder/thing_variable_builder.rs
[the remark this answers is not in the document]
alexjpwalker: I'd be tempted to turbofish the `::<Value>` or something. A single `into()` on a type that is `impl Into` may be idiomatic, two `into()` in series unqualified feels.... well.

alexjpwalker (changes requested):

rust/pattern/constraint/type_constraint.rs
dmitrii-ubskii: Kind of clashes with the naming scheme. A `FooConstraint` everywhere else is something that applies to a variable.

rust/parser/syntax_error.rs
dmitrii-ubskii: I don't think getting rid of five characters (`fmt::`) is a good enough reason to go against a convention that every other Rust project seems to follow.

rust/pattern/variable/builder/thing_variable_builder.rs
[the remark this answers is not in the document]
dmitrii-ubskii: Maybe `Value::from(string.into())`? Just so we default to using `into()` in generics where we use `impl Into<T>`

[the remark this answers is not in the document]
alexjpwalker: Sounds good to me!

rust/parser/syntax_error.rs
alexjpwalker: I had a further look into this. It looks like it is always referred to as `fmt::Display` in the Rust docs. That's probably why other Rust projects refer to it in the same way.

From a good number of crate dependencies I searched, all refer to it as `fmt::Display`. Clearly it is an accepted convention, but we do not blindly follow conventions that are not backed up by good reason.

In this case, it looks like all the methods in the `fmt` module are designed to be namespaced with `fmt`. For example, `fmt::print`, `fmt::LowerHex`. If we pick these apart (the latter in particular) the need for `fmt` prefix becomes clear: we're implementing a lower hex _formatter_, not a "lower hex".

Likewise, when we implement `Display`, we're implementing a Display formatter (ie pretty-print), and when we implement `Debug`, we're implementing a debug formatter. So in conclusion, the convention is a sensible one and we should follow it.

As for why we put in the effort to decide at all - it's weird at a glance because we constantly `#[derive(Debug)]` without a thought despite it actually being `fmt::Debug`, so `Display` should intuitively be in the same boat. But it isn't, because of the context in which `Display` is defined; a formatting library that defines many format helpers, of which `Display` is one. In terms of readability, whether it was 5 or 30 characters barely matters, it's adding a new logical block to the definition of something that we implement for _almost every_ API object we define. The conventions we decide in `typeql` will go on to decide the conventions in our other Rust projects including `typedb`.

alexjpwalker (approved):

PR #238, opened 2022-10-21
Remove error propagation and force error handling as early as possible

dmitrii-ubskii:

We create a macro emulating an unstable [try-blocks feature](https://doc.rust-lang.org/beta/unstable-book/language-features/try-blocks.html) that enables us to restrict the `Result<...>` return type to fallible function calls only. We also remove error propagation through builder functions.

Previously, in order to enable the unit tests, we returned Result from all builder functions and extended the builder functions to `Result<T, E>` as well as `T`. The new `try_!` macro allows us to use the `?` operator directly by wrapping the contents of the macro in an immediately evaluated lambda expression.

### DIFF
--- a/rust/pattern/variable/builder/concept.rs
+++ b/rust/pattern/variable/builder/concept.rs
@@ -20,24 +20,18 @@
  *
  */
 
-use crate::{pattern::*, ErrorMessage};
+use crate::pattern::*;
 
 pub trait ConceptConstrainable {
     fn constrain_is(self, is: IsConstraint) -> ConceptVariable;
 }
 
 pub trait ConceptVariableBuilder: Sized {
-    fn is(self, is: impl Into<IsConstraint>) -> Result<ConceptVariable, ErrorMessage>;
+    fn is(self, is: impl Into<IsConstraint>) -> ConceptVariable;
 }
 
 impl<U: ConceptConstrainable> ConceptVariableBuilder for U {
-    fn is(self, is: impl Into<IsConstraint>) -> Result<ConceptVariable, ErrorMessage> {
-        Ok(self.constrain_is(is.into()))
-    }
-}
-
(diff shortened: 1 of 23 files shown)


rust/util/mod.rs
```diff
@@ -52,3 +52,12 @@ macro_rules! write_joined {
...
 }
+
+#[macro_export]
+macro_rules! try_ {
+    {$($stmt:tt)*} => {
+        || -> Result<_, ErrorMessage> {
+            Ok({$($stmt)*})
+        }()
+    };
+}
```
dmitrii-ubskii, line 63: The core change: wrap the statements in a lambda that returns Result and immediately call it.

rust/pattern/variable/builder/concept.rs
```diff
@@ -20,24 +20,18 @@
...
+    fn is(self, is: impl Into<IsConstraint>) -> ConceptVariable;
 }
 
 impl<U: ConceptConstrainable> ConceptVariableBuilder for U {
-    fn is(self, is: impl Into<IsConstraint>) -> Result<ConceptVariable, ErrorMessage> {
-        Ok(self.constrain_is(is.into()))
-    }
-}
-
-impl<U: ConceptVariableBuilder> ConceptVariableBuilder for Result<U, ErrorMessage> {
```
dmitrii-ubskii, line 39: The builder traits are no longer implemented for Result.

rust/query/typeql_delete.rs
```diff
@@ -46,19 +50,3 @@ impl fmt::Display for TypeQLDelete {
         f.write_str(";")
     }
 }
-
-pub trait DeleteQueryBuilder {
```
dmitrii-ubskii, line 50: Query builders were only necessary to allow the propagation of Result.

alexjpwalker (approved):

PR #240, opened 2022-10-24
Define queries support in TypeQL Rust

dmitrii-ubskii:

We add support for schema queries, viz. `define`/`undefine`, and schema query concepts such as rules.

Implement data structures, builder functions, and parser visitors for:
- `define` query;
- `undefine` query;
- `rule`;
- `abstract`.

Misc:
- factored out common parser boilerplate;
- polished off the overridden type handling in `owns` and `plays` and fixed some bugs related to them;
- migrated from Grabl to Vaticle Factory.


rust/pattern/variable/builder/type_.rs
```diff
@@ -32,6 +33,7 @@ pub trait TypeConstrainable {
 }
 
 pub trait TypeVariableBuilder: Sized {
+    fn abstract_(self) -> TypeVariable;
```
dmitrii-ubskii, line 36: `is_abstract()` to me implies a getter, so dropping the `is` here.

rust/pattern/constraint/type_/mod.rs
```diff
@@ -37,3 +37,15 @@ pub use relates::*;
...
+use std::fmt;
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct AbstractConstraint;
+
+impl fmt::Display for AbstractConstraint {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", Abstract)
+    }
+}
```
dmitrii-ubskii, line 51: The `abstract` constraint doesn't have any payload. It's implemented as a unit struct that's wrapped in an `Option<>` in the `TypeVariable`: the idea is that the variable either has this constraint, or it doesn't.

rust/pattern/variable/type_.rs
```diff
@@ -105,7 +112,16 @@ impl fmt::Display for TypeVariable {
...
+            write_joined!(
+                f,
+                ",\n    ",
+                self.sub,
+                self.regex,
+                self.relates,
+                self.plays,
+                self.owns,
+                self.abstract_
+            )?;
```
dmitrii-ubskii, line 124: Adding one more field forced `rustfmt` to multiline this call.

rust/typeql_lang.rs
```diff
@@ -42,33 +42,50 @@ mod util;
...
-    } else {
-        Err(errors
-            .borrow()
-            .iter()
-            .map(SyntaxError::to_string)
-            .collect::<Vec<String>>()
-            .join("\n\n"))
-    }
+    parse!(visit_eof_query(eof_query(query_string)))
+}
```
dmitrii-ubskii, line 83: Split the parser boilerplate into a macro. This was nicer than splitting it off into a function because a) the way antlr-rust handles lifetimes means no elision is possible, and, consequently, b) the generic function implementation exposes a lot of antlr rust implementation details that we do not necessarily want to care about.

rust/pattern/schema/rule.rs
```diff
@@ -0,0 +1,104 @@
...
+        write!(f, "{} {}", Schema::Rule, self.label)
+    }
+}
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct Rule {
+    pub label: Label,
+    pub when: Conjunction,
+    pub then: ThingVariable,
+}
```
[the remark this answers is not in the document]
alexjpwalker: Is `RuleDeclaration` a standard name for the statement `rule:{label}`, or is there a better name? What do we call it in the TypeQL ANTLR grammar?

I ask because `RuleWhenStub` and `Rule` are pretty clear, but `RuleDeclaration` leaves one wondering what it is (until reading this comment, of course!).

rust/pattern/variable/builder/type_.rs
alexjpwalker: According to the tests, the sample usage of this is
```rs
typeql_define!(
        type_("concrete-type").sub("entity"),
        type_("abstract-type").sub("entity").abstract_()
    );
```
which does, indeed, read better than `is_abstract`.

How's it named in Java? If we're using `isAbstract` for the same method in Java, we should update that too, I think. Although, that would be a breaking API change for existing users. But it seems like a correct decision.

alexjpwalker (changes requested):

rust/pattern/variable/builder/type_.rs
dmitrii-ubskii: The changes in the Java version would propagate to e.g. TypeDB core, so I'm not touching it yet. I think that's way out of scope.

alexjpwalker: In that case, let's just raise an issue for now.

rust/pattern/schema/rule.rs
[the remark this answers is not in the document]
alexjpwalker: Oh wow, the grammar is indeed really unhelpful in this case.

In that case `RuleDeclaration` is fine. The reader will have to... well, read a bit to find out what it means.

I quite like `RuleDefinition`. That would leave us with `RuleDeclaration`, `RuleWhenStub` and `RuleDefinition`. That's pretty clear actually.

alexjpwalker (commented):
Left some comments.

rust/pattern/variable/builder/type_.rs
dmitrii-ubskii: https://github.com/vaticle/typeql/issues/241

alexjpwalker (approved):

haikalpribadi (approved):

PR #245, opened 2022-11-01
Rust code style refactor

dmitrii-ubskii:

Refactor the code to be more idiomatic: use established traits and avoid star imports.

- expand all the internal star imports;
- make conversions more idiomatic;
- add missing top-level builder and parser functions;
- misc refactoring.

### DIFF
--- a/rust/builder/mod.rs
+++ b/rust/builder/mod.rs
@@ -20,7 +20,14 @@
  *
  */
 
-use crate::{common::token::Predicate, *};
+use crate::{
+    common::token::Predicate,
+    pattern::{
+        Negation, RelationVariableBuilder, RolePlayerConstraint, RuleDeclaration, ThingVariable,
+        TypeVariable, TypeVariableBuilder, UnboundVariable, Value, ValueConstraint,
+    },
+    ErrorMessage, Pattern,
+};
 
 #[macro_export]
 macro_rules! typeql_match {
@@ -124,18 +131,16 @@ where
 }
 
 pub fn gte<T: TryInto<Value>>(value: T) -> Result<ValueConstraint, ErrorMessage>
-    where
-        ErrorMessage: From<<T as TryInto<Value>>::Error>,
+where
+    ErrorMessage: From<<T as TryInto<Value>>::Error>,
 {
     ValueConstraint::new(Predicate::Gte, value.try_into()?)
 }
 
-pub fn contains<T: Into<String>>(value: T) -> Result<ValueConstraint, ErrorMessage>
-{
+pub fn contains<T: Into<String>>(value: T) -> Result<ValueConstraint, ErrorMessage> {
     ValueConstraint::new(Predicate::Contains, Value::from(value.into()))
 }
 
-pub fn like<T: Into<String>>(value: T) -> Result<ValueConstraint, ErrorMessage>
-{
+pub fn like<T: Into<String>>(value: T) -> Result<ValueConstraint, ErrorMessage> {
     ValueConstraint::new(Predicate::Like, Value::from(value.into()))
 }
--- a/rust/builder/mod.rs
+++ b/rust/builder/mod.rs
@@ -25,7 +25,7 @@ use crate::{common::token::Predicate, *};
 #[macro_export]
 macro_rules! typeql_match {
     ($($pattern:expr),* $(,)?) => {{
-        TypeQLMatch::new(vec![$($pattern.into_pattern()),*])
+        TypeQLMatch::new(vec![$($pattern.into()),*])
     }}
 }
 
@@ -39,31 +39,31 @@ macro_rules! typeql_insert {
 #[macro_export]
 macro_rules! typeql_define {
     ($($pattern:expr),* $(,)?) => {{
-        TypeQLDefine::new(vec![$($pattern.into_pattern()),*])
+        TypeQLDefine::new(vec![$($pattern.into()),*])
     }}
 }
 
 #[macro_export]
 macro_rules! typeql_undefine {
     ($($pattern:expr),* $(,)?) => {{
-        TypeQLUndefine::new(vec![$($pattern.into_pattern()),*])
+        TypeQLUndefine::new(vec![$($pattern.into()),*])
     }}
 }
 
 #[macro_export]
 macro_rules! and {
     ($($pattern:expr),* $(,)?) => {{
-        Conjunction::new(vec![$($pattern.into_pattern()),*])
+        Conjunction::new(vec![$($pattern.into()),*])
     }}
 }
 
 #[macro_export]
 macro_rules! or {
     ($($pattern:expr),* $(,)?) => {{
-        let mut patterns = vec![$($pattern.into_pattern()),*];
+        let mut patterns = vec![$($pattern.into()),*];
         match patterns.len() {
             1 => patterns.pop().unwrap(),
-            _ => Disjunction::new(patterns).into_pattern(),
+            _ => Disjunction::new(patterns).into(),
         }
     }}
 }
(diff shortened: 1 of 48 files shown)


rust/builder/mod.rs
```diff
@@ -20,12 +20,19 @@
...
+        TypeVariable, TypeVariableBuilder, UnboundVariable, Value, ValueConstraint,
+    },
+    ErrorMessage, Pattern,
+};
 
 #[macro_export]
 macro_rules! typeql_match {
     ($($pattern:expr),* $(,)?) => {{
-        TypeQLMatch::new(vec![$($pattern.into_pattern()),*])
+        TypeQLMatch::new(vec![$($pattern.into()),*])
```
dmitrii-ubskii, line 35: Replaced ad hoc `into_foo()` with `impl From<T> for Foo` implementations.

rust/builder/mod.rs
```diff
@@ -39,31 +46,43 @@ macro_rules! typeql_insert {
...
 macro_rules! or {
     ($($pattern:expr),* $(,)?) => {{
-        Disjunction::new(vec![$($pattern.into_pattern()),*])
+        let mut patterns = vec![$($pattern.into()),*];
+        match patterns.len() {
+            1 => patterns.pop().unwrap(),
+            _ => Disjunction::new(patterns).into(),
+        }
     }}
 }
```
dmitrii-ubskii, line 76: Short-circuit `or!(X)` into `X`, as in Java.

rust/builder/mod.rs
```diff
@@ -39,31 +46,43 @@ macro_rules! typeql_insert {
...
 
+pub fn not<T: Into<Pattern>>(pattern: T) -> Negation {
+    Negation::new(pattern.into())
+}
+
+pub fn rule(name: &str) -> RuleDeclaration {
+    RuleDeclaration::from(name)
+}
+
```
dmitrii-ubskii, line 85: Reordered.

rust/parser/mod.rs
```diff
@@ -33,19 +33,28 @@ use antlr_rust::{
...
+        RuleDefinition, SubConstraint, ThingConstrainable, ThingVariable, ThingVariableBuilder,
+        Type, TypeConstrainable, TypeVariable, TypeVariableBuilder, UnboundVariable, Value,
+        ValueConstraint, Variable,
+    },
+    query::{
+        sorting, AggregateQueryBuilder, Query, Sorting, TypeQLDefine, TypeQLDelete, TypeQLInsert,
+        TypeQLMatch, TypeQLMatchAggregate, TypeQLMatchGroup, TypeQLMatchGroupAggregate,
+        TypeQLUndefine, TypeQLUpdate,
+    },
 };
```
dmitrii-ubskii, line 56: Expanded star imports.

rust/parser/mod.rs
```diff
@@ -33,19 +33,28 @@ use antlr_rust::{
...
+use typeql_grammar::typeqlrustparser::*;
```
dmitrii-ubskii, line 57: Not expanding this star import because I don't think we should care about the exact types we're importing here -- it's just all the contexts and terminal nodes.

rust/parser/mod.rs
```diff
@@ -233,7 +242,7 @@ fn visit_query_match_aggregate(
     let function = ctx.match_aggregate().unwrap();
     let match_query = visit_query_match(ctx.query_match().unwrap())?;
     Ok(match visit_aggregate_method(function.aggregate_method().unwrap()) {
-        Aggregate::Count => match_query.count(),
+        token::Aggregate::Count => match_query.count(),
```
dmitrii-ubskii, line 245: Always fully specify `token::` from now on.

rust/pattern/constraint/concept.rs
```diff
@@ -41,18 +51,12 @@ impl From<String> for IsConstraint {
...
-        Self::from(var.into_concept())
-    }
-}
-
-impl From<ConceptVariable> for IsConstraint {
-    fn from(var: ConceptVariable) -> Self {
-        Self { variable: Box::new(var) }
+        Self::new(var.into_concept())
     }
 }
```
dmitrii-ubskii, line 56: This could never be used in a builder, so it's now a private constructor instead (`new()`).

rust/query/typeql_match.rs
```diff
@@ -138,46 +137,51 @@ pub struct Filter {
...
-            },
+pub mod sorting {
+    use crate::pattern::UnboundVariable;
+    use std::fmt;
+
+    #[derive(Clone, Debug, Eq, PartialEq)]
+    pub struct OrderedVariable {
+        pub var: UnboundVariable,
+        pub order: Option<String>, // FIXME
+    }
```
dmitrii-ubskii, line 153: Allows us to specify `OrderedVariable` => `sorting::OrderedVariable`, to set it apart from all the other `FooVariable`s we have.

rust/util/mod.rs
```diff
@@ -32,6 +32,17 @@ macro_rules! enum_getter {
...
+macro_rules! enum_wrapper {
+    {$enum_name:ident $($classname:ty => $enum_value:ident),* $(,)*} => {
+        $(impl From<$classname> for $enum_name {
+            fn from(x: $classname) -> Self {
+                Self::$enum_value(x)
+            }
+        })*
+    };
+}
+
```
dmitrii-ubskii, line 45: This macro allows us to quickly generate the `From` trait implementations for enums.

rust/typeql_lang.rs
```diff
@@ -88,3 +85,23 @@ pub fn parse_queries(
...
+    parse!(visit_eof_schema_rule(eof_schema_rule(typeql_rule.trim_end())))
+}
+
+pub fn parse_variable(typeql_variable: &str) -> Result<Variable, String> {
+    parse!(visit_eof_variable(eof_variable(typeql_variable.trim_end())))
+}
+
+pub fn parse_label(typeql_label: &str) -> Result<Label, String> {
+    parse!(visit_eof_label(eof_label(typeql_label.trim_end())))
+}
```
dmitrii-ubskii, line 107: Add missing top-level parse fns.

rust/typeql_lang.rs
```diff
@@ -20,17 +20,15 @@
  *
  */
 
-#![allow(dead_code)]
```
dmitrii-ubskii, line 23: There should be no more unused private functions now. (There are, but they are all the would-be-red tests).

rust/parser/mod.rs
alexjpwalker: The reason we avoid star imports in general (in all programming languages) is so that, given any symbol in the file, it's easy to look at the import list and see where the symbol comes from.

Leaving this star import in would break that convention.

rust/util/mod.rs
alexjpwalker: Ah, is this a macro that lets us easily implement `From` to, in a way, "downcast" an enum to one of its members (or the other way, "upcasting" an enum variant to the parent enum)? I've been looking for one of these.

If so, is there not a crate already out there that provides such functionality?

alexjpwalker (changes requested):
Left some comments.

rust/parser/mod.rs
dmitrii-ubskii: The idea here is that we want to import all the contexts from the generated grammar, because those are the types we intend to visit. Listing all fifty-odd types seems needlessly noisy.

rust/util/mod.rs
dmitrii-ubskii: Not that I've found. There's `enum-as-inner` that does the same thing as the `enum_getter` macro. Either way, I hesitate adding more dependencies for something this trivial.

PR #246, opened 2022-11-02
Enable disjunctions in rules

krishnangovindraj:
Enables disjunctions in rules, and updates rule-validation accordingly.

* Updates rule-validation to allow disjunctions in rule-conditions. 
* updates other validation steps to reflect disjunctions being allowed
  - Every variable in the rule-then must be present in the trunk of the rule-when (i.e., outside negations and nested disjunctions).

### DIFF
--- a/java/common/exception/ErrorMessage.java
+++ b/java/common/exception/ErrorMessage.java
@@ -80,25 +80,25 @@ public class ErrorMessage extends com.vaticle.typedb.common.exception.ErrorMessa
     public static final ErrorMessage INVALID_RULE_WHEN_NESTED_NEGATION =
             new ErrorMessage(29, "Rule '%s' 'when' contains a nested negation.");
     public static final ErrorMessage INVALID_RULE_THEN =
-            new ErrorMessage(31, "Rule '%s' 'then' '%s': must be exactly one attribute ownership, or exactly one relation.");
+            new ErrorMessage(30, "Rule '%s' 'then' '%s': must be exactly one attribute ownership, or exactly one relation.");
     public static final ErrorMessage INVALID_RULE_THEN_HAS =
-            new ErrorMessage(32, "Rule '%s' 'then' '%s' tries to assign type '%s' to variable '%s', but this variable already had a type assigned by the rule 'when'. Try omitting this type assignment.");
+            new ErrorMessage(31, "Rule '%s' 'then' '%s' tries to assign type '%s' to variable '%s', but this variable already had a type assigned by the rule 'when'. Try omitting this type assignment.");
     public static final ErrorMessage INVALID_RULE_THEN_VARIABLES =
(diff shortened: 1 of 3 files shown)


java/common/exception/ErrorMessage.java
```diff
@@ -79,14 +79,12 @@ public class ErrorMessage extends com.vaticle.typedb.common.exception.ErrorMessa
...
             new ErrorMessage(29, "Rule '%s' 'when' contains a nested negation.");
-    public static final ErrorMessage INVALID_RULE_WHEN_CONTAINS_DISJUNCTION =
-            new ErrorMessage(30, "Rule '%s' 'when' contains a disjunction.");
     public static final ErrorMessage INVALID_RULE_THEN =
             new ErrorMessage(31, "Rule '%s' 'then' '%s': must be exactly one attribute ownership, or exactly one relation.");
     public static final ErrorMessage INVALID_RULE_THEN_HAS =
             new ErrorMessage(32, "Rule '%s' 'then' '%s' tries to assign type '%s' to variable '%s', but this variable already had a type assigned by the rule 'when'. Try omitting this type assignment.");
     public static final ErrorMessage INVALID_RULE_THEN_VARIABLES =
-            new ErrorMessage(33, "Rule '%s' 'then' variables must be present in rule 'when'.");
+            new ErrorMessage(33, "Rule '%s' 'then' variables must be present in every branch of rule 'when'.");
```
flyingsilverfin, line 87: we don't have `branch` on TypeQL. You need to use terminology local to the domain. 
The validation should actually require that every `then` variable is bound outside outside of the disjunction before normalisation:
```
$x $y $z;
{ $x has $a; } or { $y has $a }
```
we only allow x, y, z to be used in the then

krishnangovindraj: ` 'then' variables must be present in the 'when' outside of disjunctions & negations.`
OR
` 'then' variables must be present in the trunk of the 'when' `

java/common/exception/ErrorMessage.java
```diff
@@ -79,14 +79,12 @@ public class ErrorMessage extends com.vaticle.typedb.common.exception.ErrorMessa
             new ErrorMessage(28, "Rule '%s' 'when' has not been provided with any patterns.");
     public static final ErrorMessage INVALID_RULE_WHEN_NESTED_NEGATION =
             new ErrorMessage(29, "Rule '%s' 'when' contains a nested negation.");
-    public static final ErrorMessage INVALID_RULE_WHEN_CONTAINS_DISJUNCTION =
-            new ErrorMessage(30, "Rule '%s' 'when' contains a disjunction.");
     public static final ErrorMessage INVALID_RULE_THEN =
```
flyingsilverfin, line 82: we currently have the habit of bumping down all the version numbers so there are no gaps

krishnangovindraj: Updated.
But we should fix that with the rust version, at least.

flyingsilverfin (approved):

PR #249, opened 2022-11-10
Parameter and structure validation in TypeQL Rust

dmitrii-ubskii:

We implement methods that allow us to check the validity of parsed and constructed typeql queries and patterns, and that provide a report of all the errors discovered during validation.

- Implement validation routines that e.g. check that conjunctions contain a non-zero number of patterns;
- modify the visitor functions s.t. no errors are produced until the very last step;
- modify the builder functions s.t. no errors are produced at all; it is now the responsibility of whatever component accepts the build results to ensure its inputs are valid;
- the above changes allow us to compile a report of all errors during validation, as opposed to only the first encountered one.;
- backport the undefine rule check to TypeQL Java, along with the corresponding error message.


java/query/TypeQLDefinable.java
```diff
@@ -51,7 +51,13 @@ abstract class TypeQLDefinable extends TypeQLQuery {
...
         this.definables = new ArrayList<>(definables);
         for (Definable definable : definables) {
-            if (definable.isRule()) rules.add(definable.asRule());
+            if (definable.isRule()) {
+                Rule rule = definable.asRule();
+                if (rule.when() != null || rule.then() != null) {
+                    throw TypeQLException.of(ErrorMessage.INVALID_UNDEFINE_QUERY_RULE.message(rule.label()));
+                }
+                rules.add(rule);
+            }
```
dmitrii-ubskii, line 60: Backported the undefine rule check: no when or then clause allowed in undefine.

rust/pattern/variable/type_.rs
```diff
@@ -66,6 +81,29 @@ impl TypeVariable {
             || self.sub.is_some()
             || self.value_type.is_some()
     }
+
+    pub fn validate_definable(&self) -> Result<()> {
+        if self.label.is_none() {
+            Err(INVALID_DEFINE_QUERY_VARIABLE.format(&[]))?;
+        }
+        Ok(())
+    }
```
dmitrii-ubskii, line 90: Separate validation check for when the type variable is within an [un]define query.

rust/query/typeql_define.rs
```diff
@@ -34,23 +39,34 @@ pub struct TypeQLDefine {
...
-                Pattern::Rule(rule) => define.add_rule(rule),
-                Pattern::Variable(Variable::Type(var)) => define.add_definition(var),
-                _ => unreachable!(),
+    pub fn new(definables: Vec<Definable>) -> Self {
+        definables.into_iter().fold(TypeQLDefine::default(), |define, definable| match definable {
+            Definable::RuleDefinition(rule) => define.add_rule(rule),
+            Definable::TypeVariable(var) => define.add_definition(var),
+            Definable::RuleDeclaration(r) => {
+                panic!("{}", INVALID_RULE_WHEN_MISSING_PATTERNS.format(&[&r.to_string()]))
             }
```
dmitrii-ubskii, line 48: I'd quite like to split definable and undefinable into separate enums, but that clashes with the way the grammar is formulated currently.

rust/typeql_lang.rs
```diff
@@ -38,70 +30,75 @@ pub mod query;
...
         let errors = Rc::new(RefCell::new(Vec::<SyntaxError>::new()));
         parser.add_error_listener(Box::new(ErrorListener::new(input, errors.clone())));
 
-        let result = $visitor(parser.$accessor().unwrap());
+        let ast_root = parser.$accessor().unwrap();
 
         if errors.borrow().is_empty() {
-            result.map_err(|em| em.message)
+            $visitor(ast_root)
         } else {
```
dmitrii-ubskii, line 68: First check there were no syntax errors, then try to visit the root node, otherwise the tree we visit will be in an invalid state. Previously the visitors couldn't panic, so this bug was invisible.

rust/util/mod.rs
```diff
@@ -63,12 +63,3 @@ macro_rules! write_joined {
...
 }
-
-#[macro_export]
-macro_rules! try_ {
-    {$($stmt:tt)*} => {
-        || -> Result<_, ErrorMessage> {
-            Ok({$($stmt)*})
-        }()
-    };
-}
```
dmitrii-ubskii, line 74: No longer required for error handling.

rust/pattern/constraint/type_/regex.rs
```diff
@@ -20,14 +20,20 @@
...
 #[derive(Debug, Clone, Eq, PartialEq)]
 pub struct RegexConstraint {
     regex: String,
 }
 
+impl Validatable for RegexConstraint {
+    fn validate(&self) -> Result<()> {
+        Ok(()) // TODO validate regex using external crate
+    }
+}
```
dmitrii-ubskii, line 35: Not validating in this PR, as this check requires adding the regex or regex-syntax crate to dependencies. This will be done in a separate pull.

rust/query/typeql_insert.rs
```diff
@@ -35,6 +45,43 @@ impl TypeQLInsert {
...
+        )
+    }
+}
+
+fn expect_insert_in_scope_of_match(
+    match_query: &Option<TypeQLMatch>,
+    variables: &[ThingVariable],
+) -> Result<()> {
+    if let Some(match_query) = match_query {
+        let bounds = match_query.conjunction.names();
```
alexjpwalker, line 67: Can we clarify (maybe with the Core team) what is the difference between "bounds" and "scope"?

"scope" is definitely ambiguous, and everywhere we use "bounds" it feels like scope would be appropriate too...

rust/query/writable.rs
```diff
@@ -40,6 +40,13 @@ impl<const N: usize> Writable for [ThingVariable; N] {
 
 impl Writable for Vec<ThingVariable> {
     fn vars(self) -> Vec<ThingVariable> {
-        self
+        self.to_vec()
```
alexjpwalker, line 43: ?)?)?)?

alexjpwalker (changes requested):

rust/parser/syntax_error.rs
```diff
@@ -22,7 +22,10 @@
 
 use std::fmt;
 
-use crate::common::error::{ErrorMessage, SYNTAX_ERROR_DETAILED, SYNTAX_ERROR_NO_DETAILS};
+use crate::common::{
+    error,
+    error::{SYNTAX_ERROR_DETAILED, SYNTAX_ERROR_NO_DETAILS},
+};
```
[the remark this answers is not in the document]
alexjpwalker: Looks elegant to me!

alexjpwalker (commented):
Changes look good to me, but we should confirm that it's appropriate to call `Scope` "scope". It's always been called "bounds". I prefer the name "scope", I think it's clearer, but it's also overloaded, so we do need to be sure.

alexjpwalker (approved):

haikalpribadi (approved):

PR #256, opened 2022-11-24
Cucumber

dmitrii-ubskii:
**For the title of this PR:** please follow the grammatical rules of a usual publication title, without capitalisation (except for the first letter). Thus, the title should NOT CONTAIN CODE: no dots, no parentheses, no backticks, no brackets, etc. It needs to be distinctive (not detailed) and succinct (not lengthy). Details of this PR will go in the description. **For the description of this PR:** please replace every line in curly brackets ( { like this } ) with an appropriate description following the guidance. Finally, **please remove this paragraph**.

{ In the form of a paragraph (only use bullet points if strictly necessary), please describe the goal of this PR, why they are valuable to achieve, and reference the related GitHub issues. This section will be automatically compiled into the release notes, so please:
  - describe the impact of the change in this PR to the _user_ of this repository (e.g. end user, contributor, developer).
  - describe the new product behaviour in _present tense_, and the old behaviour and how it's been changed in _past tense_.
  - Use the _Royal We_: _"We"_ made changes, not _"I"_ made changes. }

{ Please explain what you implemented, why your changes are the best way to achieve the goal(s) above. Please describe every method, class and package, by explaining:
  - its responsibility, 
  - how it's expected to behave, and 
  - how it relates to the adjacent methods/classes/packages it interacts with. 

This would allow the reviewer to understand your intentions in the code much better. If you're adding new classes, make sure these explanations are also included in the class header comments. Last but not least, please reference the GitHub issues to be automatically closed, such like 'closes #number'. }


.factory/automation.yml
```diff
@@ -60,6 +60,7 @@ build:
       image: vaticle-ubuntu-22.04
       command: |
         bazel test //rust/... --test_output=errors
+        bazel test //rust/tests/behaviour/... --test_output=errors
```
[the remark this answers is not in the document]
alexjpwalker: `//rust/tests/behaviour/...` is a subset of `//rust/...`, so we shouldn't have both lines here.

To be honest, I'm not entirely sure why the `test-java` job is so specific either. It could just run `bazel test //java/...`.

rust/tests/behaviour/main.rs
```diff
@@ -0,0 +1,151 @@
...
+use cucumber::{gherkin::Step, given, then, when, World};
+
+use typeql_lang::parse_query;
+
+#[derive(Debug, Default, World)]
+pub struct TypeQLWorld;
+
+fn main() {
+    futures::executor::block_on(
+        TypeQLWorld::cucumber().filter_run("../vaticle_typedb_behaviour", |_, _, sc| {
```
alexjpwalker, line 32: Where does this path come from - and how reliable is it? i.e. would the path still be `../vaticle_typedb_behaviour` if we refactored this package and moved it somewhere else?

It would be beneficial to future readers to clarify in the code - either here or in the `BUILD` file - what this directory is and how it comes about.

rust/tests/behaviour/main.rs
```diff
@@ -0,0 +1,151 @@
...
+        }),
+    );
+}
+
+#[given("reasoning schema")]
+#[given("typeql define")]
+#[then("typeql define")]
+#[when("typeql define")]
+async fn typeql_define(_: &mut TypeQLWorld, step: &Step) {
+    let parsed = parse_query(&step.docstring.as_ref().unwrap()).unwrap();
```
alexjpwalker, line 43: Let's maybe extract these `parse_query((...).unwrap()).unwrap()` calls into a helper function - something like `extract_query_from_step` or `parse_query_from_step`

rust/tests/behaviour/main.rs
```diff
@@ -0,0 +1,151 @@
...
+
+#[given("typeql delete")]
+#[then("typeql delete")]
+#[when("typeql delete")]
+async fn typeql_delete(_: &mut TypeQLWorld, step: &Step) {
+    let parsed = parse_query(&step.docstring.as_ref().unwrap()).unwrap();
+    assert_eq!(parsed, parse_query(&parsed.to_string()).unwrap());
+}
+
+#[when("typeql update")]
```
alexjpwalker, line 72: Are we matching the usage of `given`, `when` and `then` to the keywords that they are actually used in in the `behaviour` repo?

I'm inclined to make us not have to worry about it:
```rs
macro_rules! step {
    $str:literal => {
        #[given($str)]
        #[when($str)]
        #[then($str)]
    }
}
```
... assuming that's actually a valid macro by itself.

If it's not possible to make a macro cleanly then we can just leave it as-is. There's just a risk of introducing errors that only get caught at CI time.

alexjpwalker (changes requested):

alexjpwalker (approved):

haikalpribadi (approved):

PR #258, opened 2022-12-09
Idiomatic Rust error message enum

dmitrii-ubskii:

We replace the `Message` error struct with the `TypeQLError` enum, which greatly simplifies error handling on the user side.

The `error_messages!` macro used to generate message templates has been replaced with one that generates an error enum instead. Returning an enum as the error is the established Rust approach as it greatly simplifies error handling. We also export the error generator macro so that it can be reused in e.g. typedb-client-rust.

Another minor change is adding the `group_imports = "StdExternalCrate"` to `rustfmt.toml`, a change backported from client-rust.

### DIFF
--- a/rust/BUILD
+++ b/rust/BUILD
@@ -50,7 +50,7 @@ rust_library(
     proc_macro_deps = [
         "@vaticle_dependencies//library/crates:pest_derive",
     ],
-    visibility = ["//rust:__subpackages__"]
+    visibility = ["//visibility:public"]
 )
 
 rust_test(
--- a/rust/common/error.rs
+++ b/rust/common/error.rs
@@ -20,25 +20,30 @@
  *
  */
 
-use crate::write_joined;
+use crate::{
+    common::token,
+    pattern::{Label, Pattern, Reference, ThingVariable, TypeVariable, UnboundVariable, Value},
+    write_joined,
+};
+use chrono::NaiveDateTime;
 use pest::error::{Error as PestError, LineColLocation};
-use std::{convert::Infallible, fmt};
+use std::fmt;
 
 #[derive(Debug)]
 pub struct Error {
-    messages: Vec<Message>,
+    parse_errors: Vec<ErrorMessage>,
 }
 
-impl From<Message> for Error {
-    fn from(message: Message) -> Self {
-        Self { messages: vec![message] }
+impl From<ErrorMessage> for Error {
+    fn from(parse_error: ErrorMessage) -> Self {
+        Self { parse_errors: vec![parse_error] }
     }
 }
 
-impl From<Vec<Message>> for Error {
-    fn from(messages: Vec<Message>) -> Self {
-        assert!(!messages.is_empty());
-        Self { messages }
+impl From<Vec<ErrorMessage>> for Error {
+    fn from(parse_errors: Vec<ErrorMessage>) -> Self {
+        assert!(!parse_errors.is_empty());
+        Self { parse_errors }
     }
 }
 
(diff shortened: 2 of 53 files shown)


rust/common/error.rs
```diff
@@ -140,54 +100,151 @@ macro_rules! max {
...
+        }
+    };
+    ($self:ident, $error_name:ident, $body:literal, $t1:ty, $t2:ty, $t3:ty, $t4:ty) => {
+        if let Self::$error_name(a, b, c, d) = &$self {
+            format!($body, a, b, c, d)
+        } else {
+            unreachable!()
+        }
+    };
+}
```
dmitrii-ubskii, line 135: This is the accepted way to do varargs in Rust.

rust/common/error.rs
```diff
@@ -140,54 +100,151 @@ macro_rules! max {
     ($x:literal, $($xs:literal),*) => (max($x, max!($($xs),*)));
 }
 
+macro_rules! format_message {
+    ($self:ident, $error_name:ident, $body:literal, ) => {
+        format!($body)
```
dmitrii-ubskii, line 105: `format!()` enables the argument checking. This will not compile if `$body` contains any kind of formatting that expects arguments (i.e. `{}`)

rust/common/error.rs
```diff
@@ -140,54 +100,151 @@ macro_rules! max {
...
+    };
+    ($self:ident, $error_name:ident, $body:literal, $t1:ty) => {
+        if let Self::$error_name(a) = &$self {
+            format!($body, a)
+        } else {
+            unreachable!()
+        }
```
dmitrii-ubskii, line 112: Because of Rust macro hygiene we have to resort to this double checking. In C-like macros we'd generate a full match arm, or generate bindings in the pattern and use them in the match result, but neither of those is allowed in Rust.

rust/common/error.rs
```diff
@@ -140,54 +100,151 @@ macro_rules! max {
...
-   ILLEGAL_GRAMMAR = 39: "Illegal grammar: '{}'",
-   ILLEGAL_CHAR_IN_LABEL = 40: "'{}' is not a valid Type label. Type labels must start with a letter, and may contain only letters, numbers, '-' and '_'.",
+error_messages! { TypeQLError
+    code: "TQL", type: "TypeQL Error",
+    SyntaxErrorDetailed(usize, String, String, String) =
+        3: "There is a syntax error at line {}:\n{}\n{}\n{}",
+    InvalidCasting(&'static str, &'static str, &'static str, &'static str) =
+        4: "Enum '{}::{}' does not match '{}', and cannot be unwrapped into '{}'.",
+    MissingPatterns() =
+        5: "The query has not been provided with any patterns.",
```
dmitrii-ubskii, line 191: We now return an enum variant wrapping format arguments. This cleans up the point of use: `SYNTAX_ERROR_DETAILED.format(&[&line.to_string(), ...])` becomes `TypeQLError::SyntaxErrorDetailed(line, ...)`, and also allows the compiler to verify the types of the arguments and the number of the arguments to the format string. This approach also unlocks full capabilities of built-in Rust formatting, though we aren't using it anywhere yet.

rust/common/error.rs
```diff
@@ -20,25 +20,32 @@
...
+
+use chrono::NaiveDateTime;
 use pest::error::{Error as PestError, LineColLocation};
-use std::{convert::Infallible, fmt};
+
+use crate::{
+    common::token,
+    pattern::{Label, Pattern, Reference, ThingVariable, TypeVariable, UnboundVariable, Value},
+    write_joined,
+};
```
dmitrii-ubskii, line 32: This arrangement of imports, dubbed StdExternalCrate, is a de facto Rust community standard. Enforced here to comply with client-rust formatting, where it makes a much bigger difference.

rust/BUILD
```diff
@@ -50,7 +50,7 @@ rust_library(
     proc_macro_deps = [
         "@vaticle_dependencies//library/crates:pest_derive",
     ],
-    visibility = ["//rust:__subpackages__"]
+    visibility = ["//visibility:public"]
```
dmitrii-ubskii, line 53: Make the library visible externally.

rust/parser/mod.rs
```diff
@@ -23,10 +23,14 @@
 #[cfg(test)]
 mod test;
 
+use chrono::{NaiveDate, NaiveDateTime};
```
dmitrii-ubskii, line 26: The rest of the changes in this PR are replacing the old errors with the new ones and rustfmt import and reexport reordering.

flyingsilverfin (approved):

rust/common/error/macros.rs
```diff
@@ -0,0 +1,140 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+
+#[macro_export]
+macro_rules! error_messages {
```
flyingsilverfin, line 24: should this file be called `/rust/common/error/error_message.rs`?

flyingsilverfin (approved):

PR #260, opened 2023-01-04
Introduce expressions and computed value variables

krishnangovindraj:

We introduce the ability to perform arithmetic computation and store the results in a 'value variable' - denoted by a preceding `?`. For example:
```typeql
match
  $x isa triangle, has base $b, has height $h;
  ?area = 0.5 * $b * $h;
```

Or through the java api:
```java
TypeQLMatch query = match(
        cVar("x").isa("triangle").has("base", cVar("b")).has("height", cVar("h")),
        vVar("area").assign(Expression.constant(0.5).mul(cVar("b").mul(cVar("h"))))))
);
```

Rules now also support value-variable conclusions:
```
define
rule computation-cost-dollars-to-pounds: when {
  $c isa computation, has cost-dollars $c;
  $_ isa exchange-rate, has name "dollar-pound", has rate $rate;
  ?pounds = $rate * $c;
} then {
  $c has cost-pounds ?pounds;
};
```

The expression on the right hand side of an assignment can be functions, operations, variables or constants:
* This PR implements infix operators: `+`, `-`, `*`, `/`, `^`, `%`, 
* This PR implements prefix functions: `min`, `max`, `floor`, `ceil`, `round`, `abs`
* This PR implements parantheses: `(...)` and instantiation of constants eg. `12`, `0.25`, `false`, `"abc"`, etc.

These constructs are currently defined on double and long valued Attribute instances or Value instances.

The language implements order-of-operations for infix operators in this order: `()`, `^`, `*`, `|`, `%`, `+`, `-`.

**Deprecation warnings**
* Using `=` can should no longer be used to denote value-equality. `=` now represents value-assignment, with `==` representing value equality. For the time being, concept variables `$x` will still support the old syntax, `=` and the new `==`, however expect `$x = ` to be removed from the language in future releases.

**Breaking changes**
* `var` is no longer part of the TypeQL builder API, being replaced by `cVar` to create concept variables (`$x`) and `vVar` to create a value variable (`?y`).

Grammar:
* We introduce the idea of "Value" variables, that are represented with `?x` instead of `$x`. `$x` variables are now referred to as "Concept" variables
* Replace `=` from equality test to mean assignment for value variables: `?x = `
  * `=` for predicates and comparisons becomes `==`. For `Concept` variables, we still allow `=` and `==` to mean comparison. NOTE: the old synta x`$x =` is now **deprecated** and will be removed in 3.0
  * The right hand side of `?x = ` is an `Expression`
* Expressions are a new sub-language used for computation
  * We initially support infix operators: `+`, `-`, `*`, `/`, `^`, `%`, and wrapper expression `(...)` 
  * We intially support prefix functions: `min`, `max`, `floor`, `ceil`, `round`, `abs`
  * These are initially defined for `long` and `double` value types
  * Both `Value` and `Concept` variables can be used in expressions, along with constants

Language API:
* Introduce expression builders on `UnboundValueVariable` (`cVar(x)` and `UnboundValueVariable` (`vVar(x)`), to replace `var(x)`
* Introduce `Expression`, which when statically imported can be used to build expressions: `Expression.pow(..., ...)`, along with other expression constructs
* Replace non-typed shorthands for `get`, `sort` etc: `match(...).get("x")` with only the typed versions: `match(...).get(cVar("x"))`

Pattern:
* We introduce the new paradigm that `Unbound` variables may be used as builders for expressions or constraints. However, any variable in a `Constraint` must be a `BoundVariable`
* We extract a common `Predicate` class that is held by both Thing and Value `PredicateConstraint`s
* Introduce the new subtype of `BoundVariable`: `ValueVariable`, which holds the new `ValueConstraint`s
  * `ValueConstraint` is the parent class of `ValueConstraint.Predicate` and `ValueConstraint.Assignment`
  * `ValueConstraint.Assignment` holds an `Expression`
  * `ValueVariables` always use a `Reference.Named.Value` reference type. We refactor `Reference.Referrable` to be called `Reference.Named` and place `Reference.Named.Concept` and `Reference.Named.Value` under it, and move `Label` to `Reference.Label`.
  * We introduce `UnboundValueVariable` analogously to `UnboundConceptVariable`. Both are now subtypes of `UnboundVariable`.
* We validate that concept variables (`$x`) and value variables `?x` do not re-use the same names


java/pattern/constraint/TypeConstraint.java
```diff
@@ -27,7 +27,7 @@
 import com.vaticle.typeql.lang.common.TypeQLToken;
 import com.vaticle.typeql.lang.common.exception.TypeQLException;
 import com.vaticle.typeql.lang.pattern.variable.TypeVariable;
-import com.vaticle.typeql.lang.pattern.variable.UnboundVariable;
+import com.vaticle.typeql.lang.pattern.variable.UnboundDollarVariable;
```
krishnangovindraj, line 30: Only the rename

java/pattern/constraint/ThingConstraint.java
```diff
@@ -479,22 +470,14 @@ public int hashCode() {
         }
     }
 
-    public abstract static class Value<T> extends ThingConstraint {
+    public static class Predicate<T> extends ThingConstraint {
 
-        private final TypeQLToken.Predicate predicate;
-        private final T value;
+        private final com.vaticle.typeql.lang.pattern.expression.Predicate predicate;
```
krishnangovindraj, line 475: Pull predicate out to be common.

java/pattern/expression/Predicate.java
```diff
@@ -0,0 +1,302 @@
...
+        @Override
+        public boolean isValueVariable() {
+            return true;
+        }
+
+        @Override
+        public ValueVariable asValueVariable() {
+            return this;
+        }
+    }
```
krishnangovindraj, line 301: Introduce predicates with ValueVariable on RHS.
Differentiate wrt predicates with Dollar Variable on RHS

java/pattern/variable/Reference.java
```diff
@@ -124,44 +144,106 @@ public Referable asReferable() {
...
-                throw TypeQLException.of(INVALID_VARIABLE_NAME.message(name, REGEX.toString()));
-            }
-            this.name = name;
-            this.hash = Objects.hash(this.type, this.isVisible, this.name);
-        }
-
-        @Override
-        public String name() {
-            return name;
+    public static abstract class Name extends Referable {
```
krishnangovindraj, line 147: Make Name abstract, Introduce Name.NamedDollar and Name.NamedValue

java/pattern/variable/UnboundDollarVariable.java
```diff
@@ -0,0 +1,187 @@
...
+import com.vaticle.typeql.lang.pattern.variable.builder.ConceptVariableBuilder;
+import com.vaticle.typeql.lang.pattern.variable.builder.PredicateBuilder;
+import com.vaticle.typeql.lang.pattern.variable.builder.ThingVariableBuilder;
+import com.vaticle.typeql.lang.pattern.variable.builder.TypeVariableBuilder;
+
+import java.util.Collections;
+import java.util.List;
+
+public class UnboundDollarVariable extends UnboundVariable implements
+        PredicateBuilder<ThingVariable.Attribute>,
```
krishnangovindraj, line 37: New class to differentiate b/w UnboundDollarVariable and UnboundValueVariable

java/pattern/variable/UnboundValueVariable.java
```diff
@@ -0,0 +1,93 @@
...
+
+package com.vaticle.typeql.lang.pattern.variable;
+
+import com.vaticle.typeql.lang.pattern.constraint.Constraint;
+import com.vaticle.typeql.lang.pattern.constraint.ValueConstraint;
+import com.vaticle.typeql.lang.pattern.expression.Expression;
+import com.vaticle.typeql.lang.pattern.variable.builder.PredicateBuilder;
+
+import java.util.List;
+public class UnboundValueVariable extends UnboundVariable implements PredicateBuilder<ValueVariable> {
```
krishnangovindraj, line 30: New class to differentiate b/w UnboundDollarVariable and UnboundValueVariable

java/pattern/variable/builder/ThingVariableBuilder.java
```diff
@@ -115,218 +106,29 @@ default ThingVariable.Thing iid(String iid) {
...
             return constrain(new ThingConstraint.Relation.RolePlayer(roleTypeVar, playerVar));
         }
 
         ThingVariable.Relation constrain(ThingConstraint.Relation.RolePlayer rolePlayer);
     }
 
     interface Attribute {
-
-        // Attribute value equality constraint
-
```
krishnangovindraj, line 143: Move to PredicateBuilder

java/pattern/variable/UnboundVariable.java
```diff
@@ -21,40 +21,15 @@
...
-public class UnboundVariable extends Variable implements ConceptVariableBuilder,
-                                                         TypeVariableBuilder,
-                                                         ThingVariableBuilder.Common<ThingVariable.Thing>,
-                                                         ThingVariableBuilder.Thing,
-                                                         ThingVariableBuilder.Relation,
-                                                         ThingVariableBuilder.Attribute {
+import static com.vaticle.typedb.common.util.Objects.className;
+import static com.vaticle.typeql.lang.common.exception.ErrorMessage.INVALID_CASTING;
 
+public abstract class UnboundVariable extends Variable {
```
krishnangovindraj, line 29: Make abstract base class of UnboundDollarVariable and UnboundValueVariable.
Move implemented methods to those classes.

java/pattern/expression/Predicate.java
```diff
@@ -0,0 +1,302 @@
...
+import static com.vaticle.typeql.lang.common.TypeQLToken.Predicate.SubString.LIKE;
+import static com.vaticle.typeql.lang.common.exception.ErrorMessage.INVALID_CASTING;
+import static com.vaticle.typeql.lang.common.exception.ErrorMessage.INVALID_CONSTRAINT_DATETIME_PRECISION;
+import static com.vaticle.typeql.lang.common.exception.ErrorMessage.MISSING_CONSTRAINT_PREDICATE;
+import static com.vaticle.typeql.lang.common.exception.ErrorMessage.MISSING_CONSTRAINT_VALUE;
+import static com.vaticle.typeql.lang.common.util.Strings.escapeRegex;
+import static com.vaticle.typeql.lang.common.util.Strings.quoteString;
+
+public abstract class Predicate<T> {
+
```
krishnangovindraj, line 46: Predicate class ripped out of ThingConstraint.Value

java/pattern/expression/Expression.java
```diff
@@ -0,0 +1,582 @@
...
+import java.util.Objects;
+import java.util.Set;
+
+import static com.vaticle.typedb.common.collection.Collections.list;
+import static com.vaticle.typedb.common.util.Objects.className;
+import static com.vaticle.typeql.lang.common.TypeQLToken.Char.COMMA_SPACE;
+import static com.vaticle.typeql.lang.common.exception.ErrorMessage.INVALID_CASTING;
+
+public abstract class Expression {
+
```
krishnangovindraj, line 44: Expression class recursively defines expressions.
Used (only) in ValueConstraint.Expression

java/pattern/expression/Expression.java
```diff
@@ -0,0 +1,582 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.vaticle.typeql.lang.pattern.expression;
```
krishnangovindraj, line 22: This package could be named 'subpattern' or something similar.
It contains expressions and predicates.
Could also contain aggregates.

krishnangovindraj (commented):
High level comments.

dependencies/vaticle/repositories.bzl
```diff
@@ -34,6 +34,6 @@ def vaticle_typedb_common():
 def vaticle_typedb_behaviour():
     git_repository(
         name = "vaticle_typedb_behaviour",
-        remote = "https://github.com/vaticle/typedb-behaviour",
-        commit = "9ce29c4e2ccf0d34186dbd9c9708153490732d76", # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_typedb_behaviour
+        remote = "https://github.com/krishnangovindraj/typedb-behaviour",
```
krishnangovindraj, line 37: Will change after review

java/TypeQL.java
```diff
@@ -180,172 +193,257 @@ public static ThingVariable.Relation rel(String playerVar) {
...
+
+    public static ThingConstraint.Predicate contains(String value) {
+        return new ThingConstraint.Predicate(new Predicate.String(CONTAINS, value));
+    }
+
+    public static ThingConstraint.Predicate like(String value) {
+        return new ThingConstraint.Predicate(new Predicate.String(LIKE, value));
+    }
+
+    public static abstract class Expr {
```
krishnangovindraj, line 388: New builder functions for arithmetic

java/TypeQL.java
```diff
@@ -54,7 +59,7 @@
 import static com.vaticle.typeql.lang.common.TypeQLToken.Predicate.SubString.CONTAINS;
 import static com.vaticle.typeql.lang.common.TypeQLToken.Predicate.SubString.LIKE;
 import static com.vaticle.typeql.lang.common.exception.ErrorMessage.ILLEGAL_CHAR_IN_LABEL;
-import static com.vaticle.typeql.lang.pattern.variable.UnboundVariable.hidden;
+import static com.vaticle.typeql.lang.pattern.variable.UnboundDollarVariable.hidden;
```
krishnangovindraj, line 62: Most changes in this file are a result of 
- Replacing UnboundVariable with UnboundDollarVariable
- Renaming ThingConstraint.Value to ThingConstraint.Predicate

Others are pointed out with comments

java/parser/Parser.java
```diff
@@ -760,6 +786,72 @@ public Object visitValue(TypeQLParser.ValueContext ctx) {
...
+        UnboundValueVariable ownerBuilder = getValVar(ctx.VALVAR_());
+        if (ctx.predicate() != null) {
+            return ownerBuilder.constrain(new ValueConstraint.Predicate(visitPredicate(ctx.predicate())));
+        } else if (ctx.ASSIGN() != null) {
+            return ownerBuilder.constrain(new ValueConstraint.Expression(visitExpr(ctx.expr())));
+        } else throw TypeQLException.of(ILLEGAL_STATE);
+    }
+
+    @Override
+    public Expression visitExpr(TypeQLParser.ExprContext ctx) {
```
krishnangovindraj, line 801: Expression parsing.

java/parser/Parser.java
```diff
@@ -572,11 +585,11 @@ public ThingVariable.Relation visitVariable_relation(TypeQLParser.Variable_relat
 
     @Override
     public ThingVariable.Attribute visitVariable_attribute(TypeQLParser.Variable_attributeContext ctx) {
-        UnboundVariable unbound;
+        UnboundDollarVariable unbound;
         if (ctx.VAR_() != null) unbound = getVar(ctx.VAR_());
         else unbound = hidden();
 
-        ThingVariable.Attribute attribute = unbound.constrain(visitPredicate(ctx.predicate()));
+        ThingVariable.Attribute attribute = unbound.constrain(new ThingConstraint.Predicate(visitPredicate(ctx.predicate())));
```
krishnangovindraj, line 592: Effects of making Predicate common for Things & Values

java/parser/Parser.java
```diff
@@ -675,17 +688,21 @@ public Pair<String, String> visitLabel_scoped(TypeQLParser.Label_scopedContext c
     // ATTRIBUTE OPERATION CONSTRUCTS ==========================================
 
     @Override
-    public ThingConstraint.Value<?> visitPredicate(TypeQLParser.PredicateContext ctx) {
+    public Predicate<?> visitPredicate(TypeQLParser.PredicateContext ctx) {
```
krishnangovindraj, line 691: Effect of making Predicate common b/w Things & Values

java/pattern/Conjunction.java
```diff
@@ -72,8 +73,12 @@ public Stream<BoundVariable> variables() {
         });
     }
 
-    public Stream<UnboundVariable> namedVariablesUnbound() {
-        return variables().filter(Variable::isNamed).map(v -> UnboundVariable.named(v.name())).distinct();
+    public Stream<UnboundValueVariable> namedUnboundValueVariables() {
```
krishnangovindraj, line 76: Useful to have both, since they're not usable interchangeably

java/pattern/constraint/ThingConstraint.java
```diff
@@ -479,241 +475,51 @@ public int hashCode() {
...
+        public Predicate asPredicate() {
             return this;
         }
 
-        public TypeQLToken.Predicate predicate() {
+        public com.vaticle.typeql.lang.pattern.expression.Predicate<?> predicate() {
             return predicate;
         }
 
-        public T value() {
```
krishnangovindraj, line 519: Ripped out to Predicate class

java/pattern/constraint/ValueConstraint.java
```diff
@@ -0,0 +1,160 @@
...
+import java.util.Objects;
+import java.util.Set;
+
+import static com.vaticle.typedb.common.collection.Collections.set;
+import static com.vaticle.typedb.common.util.Objects.className;
+import static com.vaticle.typeql.lang.common.TypeQLToken.Char.SPACE;
+import static com.vaticle.typeql.lang.common.exception.ErrorMessage.INVALID_CASTING;
+
+public abstract class ValueConstraint extends Constraint<BoundVariable> {
+
```
krishnangovindraj, line 38: ValueConstraints can be Expression or Predicate constraints

java/pattern/expression/Expression.java
```diff
@@ -0,0 +1,533 @@
...
+
+    protected abstract void collectVariables(Set<BoundVariable> collector);
+
+    @Override
+    public abstract boolean equals(Object o);
+
+    @Override
+    public abstract int hashCode();
+
+    public static class Operation extends Expression {
```
krishnangovindraj, line 106: Operation

java/pattern/expression/Expression.java
```diff
@@ -0,0 +1,533 @@
...
+        }
+
+        @Override
+        public int hashCode() {
+            return hash;
+        }
+    }
+
+    public static class Function extends Expression {
+        private final String symbol;
```
krishnangovindraj, line 177: Function

java/pattern/expression/Expression.java
```diff
@@ -0,0 +1,533 @@
...
+        }
+
+        @Override
+        public int hashCode() {
+            return hash;
+        }
+    }
+
+    public abstract static class Constant<T> extends Expression {
+        T value;
```
krishnangovindraj, line 324: Constants

java/pattern/expression/Expression.java
```diff
@@ -0,0 +1,533 @@
...
+            return (this.symbol.equals(that.symbol) && this.argList.equals(that.argList));
+        }
+
+        @Override
+        public int hashCode() {
+            return hash;
+        }
+    }
+
+    public static class ThingVar extends Expression {
```
krishnangovindraj, line 229: Variables

java/pattern/expression/Expression.java
```diff
@@ -0,0 +1,533 @@
...
+            }
+
+            @Override
+            public DateTime asDateTime() {
+                return this;
+            }
+        }
+    }
+
+    public static class Bracketed extends Expression {
```
krishnangovindraj, line 487: Bracketed

java/pattern/variable/Reference.java
```diff
@@ -36,15 +36,19 @@ public abstract class Reference {
     final Type type;
     final boolean isVisible;
 
-    enum Type {NAME, LABEL, ANONYMOUS}
+    enum Type {NAMED_DOLLAR, ANONYMOUS, NAMED_VALUE, LABEL}
```
krishnangovindraj, line 39: Differentiate NAME into NAMED_VALUE and NAMED_DOLLAR, and classes below.

java/query/TypeQLWritable.java
```diff
@@ -70,14 +63,6 @@ abstract static class InsertOrDelete extends TypeQLWritable {
             this.hash = Objects.hash(this.command, this.match, this.variables);
         }
 
-        public List<UnboundVariable> namedVariablesUnbound() {
```
krishnangovindraj, line 73: Moved to subclasses.

java/query/builder/Aggregatable.java
```diff
@@ -31,50 +31,26 @@ default T count() {
         return aggregate(TypeQLToken.Aggregate.Method.COUNT, null);
     }
 
-    default T max(String var) {
```
krishnangovindraj, line 34: Disable string interface since "x" could now refer to "$x" or "?x"

java/query/builder/Sortable.java
```diff
@@ -39,34 +39,29 @@
 
 public interface Sortable<S, O, L> {
 
-    default S sort(String var, String... vars) {
```
krishnangovindraj, line 42: Disable string interface since "x" could now refer to "$x" or "?x"

krishnangovindraj (commented):
More self-review comments.

grammar/TypeQL.g4
```diff
@@ -152,16 +161,28 @@ attributes            :   attribute ( ',' attribute )* ;
...
                       ;
 predicate_equality    :   EQ | NEQ | GT | GTE | LT | LTE ;
 predicate_substring   :   CONTAINS | LIKE ;
 
-predicate_value       :   value | VAR_  ;
+predicate_value       :   value | VAR_  | VALVAR_ ;
+
+// ARITHMETIC EXPRESSION CONSTRUCTS ============================================
+
+expr                      :  VAR_   | VALVAR_
```
krishnangovindraj, line 177: TODO: Try to make precedence and associativity explicit in the grammar, rather than rely on parser implementation.

grammar/TypeQL.g4
```diff
@@ -277,10 +298,17 @@ DATETIME_       : DATE_FRAGMENT_ 'T' TIME_              ;
 VAR_            : VAR_ANONYMOUS_ | VAR_NAMED_ ;
 VAR_ANONYMOUS_  : '$_' ;
 VAR_NAMED_      : '$' [a-zA-Z0-9][a-zA-Z0-9_-]* ;
+VALVAR_         : '?' [a-zA-Z0-9][a-zA-Z0-9_-]* ;
 IID_            : '0x'[0-9a-f]+ ;
 LABEL_          : TYPE_CHAR_H_ TYPE_CHAR_T_* ;
 LABEL_SCOPED_   : LABEL_ ':' LABEL_ ;
 
+// ARITHMETIC
+POW                 : '^'         ;
```
krishnangovindraj, line 307: TODO: Introduce modulo

java/parser/Parser.java
```diff
@@ -760,6 +786,72 @@ public Object visitValue(TypeQLParser.ValueContext ctx) {
...
+        if (ctx.predicate() != null) {
+            return ownerBuilder.constrain(new ValueConstraint.Predicate(visitPredicate(ctx.predicate())));
+        } else if (ctx.ASSIGN() != null) {
+            return ownerBuilder.constrain(new ValueConstraint.Expression(visitExpr(ctx.expr())));
+        } else throw TypeQLException.of(ILLEGAL_STATE);
+    }
+
+    @Override
+    public Expression visitExpr(TypeQLParser.ExprContext ctx) {
+        if (ctx.POW() != null) {
```
krishnangovindraj, line 802: TODO: Try to treat operators as functions.

grammar/TypeQL.g4
```diff
@@ -152,16 +161,28 @@ attributes            :   attribute ( ',' attribute )* ;
...
-predicate             :   value
+predicate             :   value |   VALVAR_   |
                       |   predicate_equality   predicate_value
                       |   predicate_substring  STRING_
                       ;
 predicate_equality    :   EQ | NEQ | GT | GTE | LT | LTE ;
 predicate_substring   :   CONTAINS | LIKE ;
 
-predicate_value       :   value | VAR_  ;
+predicate_value       :   value | VAR_  | VALVAR_ ;
```
krishnangovindraj, line 173: Move VALVAR_ to attribute has rule

krishnangovindraj (commented):
Grammar review notes

grammar/TypeQL.g4
```diff
@@ -277,10 +298,17 @@ DATETIME_       : DATE_FRAGMENT_ 'T' TIME_              ;
...
 VAR_ANONYMOUS_  : '$_' ;
 VAR_NAMED_      : '$' [a-zA-Z0-9][a-zA-Z0-9_-]* ;
+VALVAR_         : '?' [a-zA-Z0-9][a-zA-Z0-9_-]* ;
 IID_            : '0x'[0-9a-f]+ ;
 LABEL_          : TYPE_CHAR_H_ TYPE_CHAR_T_* ;
 LABEL_SCOPED_   : LABEL_ ':' LABEL_ ;
 
+// ARITHMETIC
+POW                 : '^'         ;
+DIV                 : '/'         ;     TIMES               : '*'         ;
```
krishnangovindraj, line 308: Move = to value keywords

grammar/TypeQL.g4
```diff
@@ -102,7 +102,11 @@ pattern_negation      :   NOT '{' patterns '}'                        ;
 
 pattern_variable      :   variable_concept
                       |   variable_type
-                      |   variable_thing_any   ;
+                      |   variable_thing_any
+                      |   variable_value
+                      ;
+
+either_var            :   VAR_            |   VALVAR_         ;
```
krishnangovindraj, line 109: VAR_ should be VAR_NAMED 
and either_var should go with the variables to as a non-terminal named NAME_OR_VALUE

java/pattern/variable/UnboundValueVariable.java
```diff
@@ -0,0 +1,95 @@
...
+
+import com.vaticle.typeql.lang.pattern.constraint.Constraint;
+import com.vaticle.typeql.lang.pattern.constraint.ValueConstraint;
+import com.vaticle.typeql.lang.pattern.expression.Expression;
+import com.vaticle.typeql.lang.pattern.variable.builder.PredicateBuilder;
+
+import java.util.Collections;
+import java.util.List;
+
+public class UnboundValueVariable extends UnboundVariable implements PredicateBuilder<ValueVariable> {
```
krishnangovindraj, line 32: Explicitly introduce ValueVariableBuilder interface, move the required functions and make this implement it.

java/pattern/expression/Expression.java
```diff
@@ -0,0 +1,533 @@
+/*
```
krishnangovindraj, line 1: Move back into ValueConstraint
Rename ValueConstraint.Expression to ValueConstraint.Assign

java/pattern/expression/Predicate.java
```diff
@@ -0,0 +1,302 @@
+/*
```
krishnangovindraj, line 1: Move a level up and delete the package.

java/query/TypeQLMatch.java
```diff
@@ -463,10 +476,10 @@ public int hashCode() {
     public static class Group extends TypeQLQuery implements Aggregatable<Group.Aggregate> {
 
         private final TypeQLMatch query;
-        private final UnboundVariable var;
+        private final UnboundDollarVariable var;
         private final int hash;
 
-        public Group(TypeQLMatch query, UnboundVariable var) {
+        public Group(TypeQLMatch query, UnboundDollarVariable var) {
```
krishnangovindraj, line 482: Enable grouping on ValueVariables

krishnangovindraj (commented):
more review notes

grammar/TypeQL.g4
krishnangovindraj: Exponentiation must be right-associative

grammar/TypeQL.g4
krishnangovindraj: Revisit.
That causes us to either 
* handle it as something that isn't an == predicate OR
* write the rule conclusion contains a ValueVariable as `$x has attr_type == ?v`

java/pattern/constraint/ValueConstraint.java
```diff
@@ -0,0 +1,663 @@
...
+        return false;
+    }
+
+    public Predicate asPredicate() {
+        throw TypeQLException.of(INVALID_CASTING.message(className(this.getClass()), className(Predicate.class)));
+    }
+
+    public Assignment asExpression() {
+        throw TypeQLException.of(INVALID_CASTING.message(className(this.getClass()), className(Assignment.class)));
+    }
```
flyingsilverfin, line 68: `asAssignment`

java/pattern/constraint/ValueConstraint.java
```diff
@@ -0,0 +1,663 @@
...
+        private final int hash;
+
+        public Predicate(com.vaticle.typeql.lang.pattern.Predicate<?> predicate) {
+            this.predicate = predicate;
+            this.hash = Objects.hash(Predicate.class, this.predicate);
+        }
+
+        @Override
+        public Set<BoundVariable> variables() {
+            return set();
```
flyingsilverfin, line 82: is this really an empty set?

java/pattern/constraint/ValueConstraint.java
```diff
@@ -0,0 +1,663 @@
...
+        public boolean isPredicate() {
+            return true;
+        }
+
+        @Override
+        public Predicate asPredicate() {
+            return this;
+        }
+
+        public com.vaticle.typeql.lang.pattern.Predicate<?> predicate() {
```
flyingsilverfin, line 95: move above is/as

java/pattern/constraint/ValueConstraint.java
```diff
@@ -0,0 +1,663 @@
...
+
+            public boolean isBracketed() {
+                return false;
+            }
+
+            public boolean isThingVar() {
+                return false;
+            }
+
+            public boolean isValVar() {
```
flyingsilverfin, line 183: use `valueVar` throughout

java/pattern/constraint/ValueConstraint.java
```diff
@@ -0,0 +1,663 @@
...
+
+    public static class Assignment extends ValueConstraint {
+        private final Expression expression;
+        private final Set<BoundVariable> inputs;
+        private final int hash;
+
+        public Assignment(Expression expression) {
+            this.expression = expression;
+            this.inputs = new HashSet<>();
+            this.inputs.addAll(expression.variables());
```
flyingsilverfin, line 126: this can just be `this.inputs = new HashSet<>(expression.variables())`

java/pattern/constraint/ValueConstraint.java
```diff
@@ -0,0 +1,663 @@
...
+
+            @Override
+            public abstract boolean equals(Object o);
+
+            @Override
+            public abstract int hashCode();
+
+            public static class Operation extends Expression {
+
+                public enum OP {
```
flyingsilverfin, line 234: our enums are not usually caps, just regular Class-style notation (pascal-case)

java/pattern/constraint/ValueConstraint.java
```diff
@@ -0,0 +1,663 @@
...
+                }
+
+                @Override
+                public Function asFunction() {
+                    return this;
+                }
+
+                @Override
+                protected void collectVariables(Set<BoundVariable> collector) {
+                    argList.forEach(arg -> arg.collectVariables(collector));
```
flyingsilverfin, line 335: i think you can just call it `args` in here, since it's gotta be a list and we don't tend to put the data structure type in the variable name

java/pattern/constraint/ValueConstraint.java
```diff
@@ -0,0 +1,663 @@
...
+                    return this.variable.equals(that.variable);
+                }
+
+                @Override
+                public int hashCode() {
+                    return hash;
+                }
+            }
+
+            public abstract static class Constant<T> extends Expression {
```
flyingsilverfin, line 451: this should be called a `Value`, which also matches the grammar. Update the Expression builder too

java/pattern/variable/Reference.java
```diff
@@ -29,22 +29,29 @@
 
 import static com.vaticle.typedb.common.util.Objects.className;
 import static com.vaticle.typeql.lang.common.exception.ErrorMessage.INVALID_CASTING;
-import static com.vaticle.typeql.lang.common.exception.ErrorMessage.INVALID_VARIABLE_NAME;
 
 public abstract class Reference {
```
flyingsilverfin, line 33: what would this look like if we introduced `Reference.Value` and nested the rest of this under `Reference.Concept` -- then introduce `Named` as an interface?

flyingsilverfin: or something like this - to help the implementation in the server

flyingsilverfin (changes requested):

java/TypeQL.java
```diff
@@ -180,172 +193,257 @@ public static ThingVariable.Relation rel(String playerVar) {
...
+            return new ValueConstraint.Assignment.Expression.Operation(ValueConstraint.Assignment.Expression.Operation.OP.TIMES, a, b);
+        }
+
+        public static ValueConstraint.Assignment.Expression.Operation op_div(ValueConstraint.Assignment.Expression a, ValueConstraint.Assignment.Expression b) {
+            return new ValueConstraint.Assignment.Expression.Operation(ValueConstraint.Assignment.Expression.Operation.OP.DIV, a, b);
+        }
+
+        public static ValueConstraint.Assignment.Expression.Operation op_pow(ValueConstraint.Assignment.Expression a, ValueConstraint.Assignment.Expression b) {
+            return new ValueConstraint.Assignment.Expression.Operation(ValueConstraint.Assignment.Expression.Operation.OP.POW, a, b);
+        }
```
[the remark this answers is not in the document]
krishnangovindraj: Isn't this the more intuitive ordering?

java/TypeQL.java
```diff
@@ -180,172 +193,257 @@ public static ThingVariable.Relation rel(String playerVar) {
...
+            return new ValueConstraint.Assignment.Expression.Constant.Double(value);
+        }
+
+        public static ValueConstraint.Assignment.Expression.Constant.String constant(String value) {
+            return new ValueConstraint.Assignment.Expression.Constant.String(value);
+        }
+
+        public static ValueConstraint.Assignment.Expression.Constant.DateTime constant(LocalDateTime value) {
+            return new ValueConstraint.Assignment.Expression.Constant.DateTime(value);
+        }
```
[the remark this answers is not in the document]
krishnangovindraj: Could do. Combinatorial number of them, but convenient

java/common/TypeQLToken.java
```diff
@@ -408,4 +409,35 @@ public static Aggregate.Method of(String value) {
         }
     }
 
+    public enum Arithmetic {
```
flyingsilverfin, line 412: Assignment?

java/parser/Parser.java
```diff
@@ -513,6 +524,18 @@ public TypeVariable visitVariable_type(TypeQLParser.Variable_typeContext ctx) {
         return type;
     }
 
+    // VALUE VARIABLES =========================================================
+
+    @Override
+    public ValueVariable visitVariable_value(TypeQLParser.Variable_valueContext ctx) {
+        UnboundValueVariable ownerBuilder = getVarValue(ctx.VAR_VALUE_());
```
flyingsilverfin, line 531: is `builder` accurate since it is also a `Variable` now?

flyingsilverfin: other places call the unbound variable different things - let's be consistent

java/parser/Parser.java
```diff
@@ -672,10 +700,85 @@ public Pair<String, String> visitLabel_scoped(TypeQLParser.Label_scopedContext c
         return pair(scopedLabel[0], scopedLabel[1]);
     }
 
-    // ATTRIBUTE OPERATION CONSTRUCTS ==========================================
+    // ARITHMETIC EXPRESSIONS ==================================================
```
flyingsilverfin, line 703: i think the category is `Assignment`?

flyingsilverfin: match the grammar comments

java/parser/Parser.java
```diff
@@ -718,6 +824,12 @@ public String getRegex(TerminalNode string) {
         return unescapeRegex(unquoteString(string));
     }
 
+    public UnboundVariable getVarAny(TypeQLParser.Modifier_varContext ctx) {
```
flyingsilverfin, line 827: this is not an overridden method?

java/pattern/constraint/ThingConstraint.java
```diff
@@ -445,7 +436,12 @@ private Has(@Nullable TypeVariable type, ThingVariable<?> attribute) {
 
         @Override
         public Set<BoundVariable> variables() {
-            return set(attribute);
+            Set<BoundVariable> variables = new HashSet<>();
+            variables.add(attribute);
+            if (attribute.predicate().map(p -> p.predicate.predicate().equals(TypeQLToken.Predicate.Equality.EQ)).orElse(false)) {
```
flyingsilverfin, line 441: `Predicate` class has a `variables()` method right?

java/pattern/variable/BoundVariable.java
```diff
@@ -42,7 +41,7 @@ public abstract class BoundVariable extends Variable implements Conjunctable {
 
     @Override
     public void validateIsBoundedBy(Set<UnboundVariable> bounds) {
-        if (Stream.concat(Stream.of(this), variables()).noneMatch(v -> bounds.contains(v.toUnbound()))) {
+        if (Stream.concat(Stream.of(this), variables()).map(BoundVariable::toUnbound).noneMatch(v -> bounds.contains(v))) {
```
flyingsilverfin, line 44: this line (and other validations elsewhere) no longer need to change right? you can just use the old code?

java/pattern/variable/UnboundValueVariable.java
```diff
@@ -0,0 +1,86 @@
...
+
+import com.vaticle.typeql.lang.pattern.constraint.Constraint;
+import com.vaticle.typeql.lang.pattern.constraint.ValueConstraint;
+import com.vaticle.typeql.lang.pattern.variable.builder.ValueVariableBuilder;
+
+import java.util.Collections;
+import java.util.List;
+
+public class UnboundValueVariable extends UnboundVariable implements ValueVariableBuilder {
+    UnboundValueVariable(Reference reference) {
```
flyingsilverfin, line 32: would this look better if we had `Reference.Value` as a subtype, and `Reference.Concept` for the rest? Is there much benefit?

java/pattern/variable/Variable.java
```diff
@@ -94,7 +98,7 @@ public Reference reference() {
     }
 
     public boolean isNamed() {
-        return reference.isName();
+        return reference.isName() || reference().isValue();
```
flyingsilverfin, line 101: this is a bit awkward, we should discuss

java/pattern/variable/builder/ThingVariableBuilder.java
```diff
@@ -115,218 +113,29 @@ default ThingVariable.Thing iid(String iid) {
...
-        default ThingVariable.Attribute contains(String value) {
-            return constrain(new ThingConstraint.Value.String(CONTAINS, value));
-        }
-
-        default ThingVariable.Attribute like(String regex) {
-            return constrain(new ThingConstraint.Value.String(LIKE, regex));
-        }
-
-        ThingVariable.Attribute constrain(ThingConstraint.Value<?> constraint);
+    interface Attribute extends PredicateBuilder<ThingVariable.Attribute> {
```
flyingsilverfin, line 138: missing newline

java/pattern/variable/builder/ValueVariableBuilder.java
```diff
@@ -0,0 +1,20 @@
+package com.vaticle.typeql.lang.pattern.variable.builder;
+
+import com.vaticle.typeql.lang.pattern.constraint.Predicate;
+import com.vaticle.typeql.lang.pattern.constraint.ValueConstraint;
+import com.vaticle.typeql.lang.pattern.variable.ValueVariable;
+
+public interface ValueVariableBuilder extends PredicateBuilder<ValueVariable> {
```
flyingsilverfin, line 7: if `ThingVariableBuilder` has predicate builder methods on it, why doesn't `ValueVariableBuilder` as well?

java/TypeQL.java
```diff
@@ -180,172 +193,257 @@ public static ThingVariable.Relation rel(String playerVar) {
...
+        return new ThingConstraint.Predicate(new Predicate.ValueVariable(LTE, variable.toValue()));
+    }
+
+    public static ThingConstraint.Predicate contains(String value) {
+        return new ThingConstraint.Predicate(new Predicate.String(CONTAINS, value));
+    }
+
+    public static ThingConstraint.Predicate like(String value) {
+        return new ThingConstraint.Predicate(new Predicate.String(LIKE, value));
+    }
```
flyingsilverfin, line 386: should we make an `assign(Expression)` helper on here?

java/parser/test/ParserTest.java
```diff
@@ -408,13 +410,109 @@ public void testLongPredicateQuery() {
...
+                "(commodity: $x, qty: $q) isa order;\n" +
+                "?net = $p * $q;\n" +
+                "?gross = ?net * ( 1.0 + 0.21 );";
+        TypeQLMatch parsed = TypeQL.parseQuery(query).asMatch();
+        TypeQLMatch expected = match(
+                var("x").isa("commodity").has("price", var("p")),
+                rel("commodity", "x").rel("qty", "q").isa("order"),
+                valueVar("net").assign(Expression.times(Expression.thingVar(var("p")), Expression.thingVar(var("q")))),
+                valueVar("gross").assign(Expression.times(
+                        Expression.valueVar(valueVar("net")),
```
flyingsilverfin, line 439: let's look further to see if we can get rid of the explicit `Expression` wrapper around a variable of either value or concept variants

java/pattern/BUILD
```diff
@@ -66,9 +67,10 @@ checkstyle_test(
     include = glob([
         "*",
         "constraint/*.java",
+        "expression/*.java",
+        "schema/*.java",
         "variable/*.java",
-        "variable/builder/*.java",
-        "schema/*.java"
+        "variable/builder/*.java"
```
flyingsilverfin, line 73: nice, good catch and rearrange!

flyingsilverfin (changes requested):

java/common/TypeQLToken.java
krishnangovindraj: Expression maybe?

java/parser/Parser.java
krishnangovindraj: No
Only the visit_ ones are

java/pattern/variable/builder/ValueVariableBuilder.java
krishnangovindraj: ThingVariable.Attribute does
I'm not sure the root ThingVariable class has any  methods or how any of that works

java/query/builder/Aggregatable.java
krishnangovindraj: In a lot of places I just default it to the concept variable. I can go about fixing that

java/TypeQL.java
krishnangovindraj: Could do
But assigns in such a context are actually equality predicates

java/pattern/variable/BoundVariable.java
```diff
@@ -42,7 +41,7 @@ public abstract class BoundVariable extends Variable implements Conjunctable {
 
     @Override
     public void validateIsBoundedBy(Set<UnboundVariable> bounds) {
-        if (Stream.concat(Stream.of(this), variables()).noneMatch(v -> bounds.contains(v.toUnbound()))) {
+        if (Stream.concat(Stream.of(this), variables()).map(BoundVariable::toUnbound).filter(UnboundVariable::isConceptVariable).noneMatch(v -> bounds.contains(v))) {
```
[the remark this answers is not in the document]
krishnangovindraj: Updated to allow.
Initially, I didn't think value variables restricted the search-space, so this shouldn't have been allowed. But it is true that value variables can be used in predicates, and that's reason enough to allow.

java/pattern/constraint/ValueConstraint.java
krishnangovindraj: It's recursive though. And collector makes it a bit more explicit that it's an 'out' variable.
Calling it args sounds like it's an input argument?

java/pattern/constraint/ValueConstraint.java
krishnangovindraj: Changed the grammar to Constant, since that feels more correct.

java/pattern/variable/BoundVariable.java
krishnangovindraj: Yes

java/pattern/constraint/ThingConstraint.java
krishnangovindraj: Yes. Using that directly now.

java/pattern/variable/ValueVariable.java
```diff
@@ -0,0 +1,93 @@
...
+
+import java.util.List;
+import java.util.Objects;
+
+import static com.vaticle.typedb.common.collection.Collections.list;
+import static com.vaticle.typeql.lang.common.TypeQLToken.Char.SPACE;
+import static com.vaticle.typeql.lang.common.exception.ErrorMessage.ILLEGAL_CONSTRAINT_REPETITION;
+
+public class ValueVariable extends BoundVariable {
+    private ValueConstraint constraint;
```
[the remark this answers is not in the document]
krishnangovindraj: Done

java/pattern/constraint/ValueConstraint.java
krishnangovindraj: Thanks, Good catch.
Must have slipped through the refactoring when I ripped the predicate out.

java/parser/test/ParserTest.java
krishnangovindraj: Made both implement `ExpressionBuilder<T extends Expression>` with a method `T toExpression`
This lets us write `Expression.plus(var("x"), valueVar("y"))`

java/pattern/variable/Variable.java
flyingsilverfin: Let's try to remove `Label` from `Referable`, and rename this method to `isReferable`

you should try to see if this is easily propagated to core before we confirm the change. Plus it may not slide in Haikal's review

flyingsilverfin (changes requested):

grammar/TypeQL.g4
```diff
@@ -161,7 +169,24 @@ predicate             :   value
 predicate_equality    :   EQ | NEQ | GT | GTE | LT | LTE ;
 predicate_substring   :   CONTAINS | LIKE ;
 
-predicate_value       :   value | VAR_  ;
+predicate_value       :   value | VAR_CONCEPT_  | VAR_VALUE_ ;
+
+// EXPRESSION CONSTRUCTS ============================================
```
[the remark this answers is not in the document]
krishnangovindraj: As in?
I'd keep the expressions separate from the assignments.

java/TypeQL.java
krishnangovindraj: I see what you mean, But how would you use it?
I can only think of :`valueVar("x").assign(assign(Expression.plus(...)))`
The introduction of the static `assign` doesn't remove the methods returning an Expression.

krishnangovindraj: It will be tricky to be perfectly congruent to existing code because everything we have so far is chainable ((?)because it's fully left-associative (?)). Arithmetic isn't.

java/pattern/constraint/ValueConstraint.java
krishnangovindraj: Oh you meant the local data structure. Sorry I misunderstood. I'll make it args.

krishnangovindraj (commented):
Flag FUNC_ID_ terminal

grammar/TypeQL.g4
```diff
@@ -161,7 +169,24 @@ predicate             :   value
...
+expr                  :  expr_op    |   expr_base;
+
+expr_op               :  <assoc=right> expr_op POW expr_op                      // exponentiation is right-associative
+                      |  expr_op  (TIMES | DIV | MOD)  expr_op
+                      |  expr_op  (PLUS  | MINUS) expr_op
+                      |  expr_base
+                      ;
+expr_base             :  LPAREN expr RPAREN
+                      |  VAR_CONCEPT_   | VAR_VALUE_
+                      |  func           | constant
```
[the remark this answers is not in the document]
flyingsilverfin: let's use:
```
expr_func  : func_name (...)
func_name : EXPR_FUNC_NAME | MIN | MAX

// leave a comment here to indicate this is not the complete list
EXPR_FUNC_NAME : ...
```

java/query/TypeQLMatch.java
```diff
@@ -219,7 +220,7 @@ public Aggregate aggregate(TypeQLToken.Aggregate.Method method, UnboundVariable
     }
 
     public Group group(String var) {
-        return group(UnboundVariable.named(var));
+        return group(UnboundConceptVariable.named(var));
     }
```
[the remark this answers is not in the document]
flyingsilverfin: let's delete these in 1 commit that is easily reversible if we need to revert it

flyingsilverfin (approved):

grammar/TypeQL.g4
```diff
@@ -267,20 +306,21 @@ TRUE            : 'true'        ;
 FALSE           : 'false'       ;
 STRING_         : '"'  (~["\\] | ESCAPE_SEQ_ )* '"'
                 | '\'' (~['\\] | ESCAPE_SEQ_ )* '\''    ;
-LONG_           : ('+' | '-')? [0-9]+                   ;
```
krishnangovindraj, line 270: @flyingsilverfin , I had to remove the unary operators from the token because `?x = 5 +2` would not parse properly.

grammar/TypeQL.g4
```diff
@@ -187,15 +212,18 @@ type_native           :   THING           |   ENTITY          |   ATTRIBUTE
...
                       |   BOOLEAN         |   DATETIME        ;
-value                 :   STRING_         |   LONG_           |   DOUBLE_
-                      |   BOOLEAN_        |   DATE_           |   DATETIME_     ;
-regex                 :   STRING_         ;
+value                 :   STRING_         |   BOOLEAN_
+                      |   DATE_           |   DATETIME_
+                      |   signed_long     |   signed_double   ;
+
+signed_long           :   (PLUS | MINUS)?  LONG_              ;
+signed_double         :   (PLUS | MINUS)?  DOUBLE_            ;
```
krishnangovindraj, line 220: And move the signs here.

grammar/TypeQL.g4
```diff
@@ -161,18 +167,37 @@ predicate             :   value
...
 schema_rule           :   RULE label
                       |   RULE label ':' WHEN '{' patterns '}' THEN '{' variable_thing_any ';' '}' ;
 
-// TYPE, LABEL AND IDENTIFIER CONSTRUCTS =======================================
+// VARIABLE, TYPE, LABEL AND IDENTIFIER CONSTRUCTS ==============================
 
-type_any              :   type_scoped   | type          | VAR_          ;
-type_scoped           :   label_scoped                  | VAR_          ;
-type                  :   label                         | VAR_          ;       // A type can be a label or variable
+var_referable         :   VAR_CONCEPT_  | VAR_VALUE_                            ;
```
flyingsilverfin, line 196: can you put this by the other VAR_CONCEPT_ so I can show haikal more easily?

grammar/TypeQL.g4
```diff
@@ -161,24 +168,40 @@ predicate             :   value
...
+// EXPRESSION CONSTRUCTS =======================================================
 
-// SCHEMA CONSTRUCT =============================================================
+expression                  :   expression_operation    |   expression_structure        ;
+expression_operation        :   <assoc=right> expression_operation POW expression_operation         // exponentiation is right-associative
+                            |   expression_operation  (TIMES | DIV | MOD)  expression_operation
+                            |   expression_operation  (PLUS  | MINUS) expression_operation
+                            |   expression_structure
+                            ;
+expression_structure        :   '(' expression ')'
```
krishnangovindraj, line 181: I don't think expression_structure is the right name for this. @flyingsilverfin

java/common/TypeQLToken.java
```diff
@@ -408,4 +409,35 @@ public static Aggregate.Method of(String value) {
         }
     }
 
+    public enum Expression {
```
flyingsilverfin, line 412: Should we have functions as tokens too?

grammar/TypeQL.g4
flyingsilverfin: I don't think it's quite right either, but let's keep it for now. I just noticed that both `expression` and `expression_operation` fall through to `expression_structure` - is this ambiguous? Can you get rid of `expression_structure` inside `expression_operation`?

java/parser/Parser.java
```diff
@@ -513,6 +543,18 @@ public TypeVariable visitVariable_type(TypeQLParser.Variable_typeContext ctx) {
...
 
+    // VALUE VARIABLES =========================================================
+
+    @Override
+    public ValueVariable visitVariable_value(TypeQLParser.Variable_valueContext ctx) {
+        UnboundValueVariable unbound = getVarValue(ctx.VAR_VALUE_());
+        if (ctx.predicate() != null) {
+            return unbound.constrain(new ValueConstraint.Predicate(visitPredicate(ctx.predicate())));
+        } else if (ctx.ASSIGN() != null) {
+            return unbound.constrain(new ValueConstraint.Assignment(this.visitExpression(ctx.expression())));
```
flyingsilverfin, line 554: don't need `this.`

java/query/test/TypeQLQueryTest.java
```diff
@@ -36,63 +37,62 @@
...
-import static com.vaticle.typeql.lang.TypeQL.var;
 import static org.junit.Assert.assertEquals;
 
 // TODO: This test should be split into one TypeQL query test class each
 public class TypeQLQueryTest {
 
     @Test
     public void testSimpleGetQueryToString() {
-        assertSameStringRepresentation(match(var("x").isa("movie").has("title", "Godfather")));
+        assertSameStringRepresentation(match(TypeQL.cVar("x").isa("movie").has("title", "Godfather")));
```
flyingsilverfin, line 47: I think you should still static-import `cVar` and `vVar` in all the tests

flyingsilverfin (changes requested):
Just left a few independent comments. The grammar now looks nice and clean, good job!

The big remaining thing to do is do summarise the language/API changes in the PR "Goal" so that that gets published onto the relase notes.

Eg. people can see a quick summary of the API changes: Introduction of `?var` and new terminology. `var` -> `cVar`. Introduction of `vVar`. Change of `=` to `==`, introduction of `=`.  Basic example of how to use assignments/expressions in TypeQL and the equivalent using the TypeQL programmatic API.

I think having this as something linkable will save us a lot of headaches when people ask for API changes on release!

grammar/TypeQL.g4
krishnangovindraj: Yes. That also gives us a redundant expression : expression_operation
So we can collapse `expression`, `expression_operation` and `expression_structure` the whole thing into a single rule. Shall we?

flyingsilverfin: Ok we might as well

java/pattern/variable/UnboundValueVariable.java
```diff
@@ -0,0 +1,99 @@
...
+import com.vaticle.typeql.lang.pattern.variable.builder.PredicateBuilder;
+import com.vaticle.typeql.lang.pattern.variable.builder.ValueVariableBuilder;
+
+import java.util.Collections;
+import java.util.List;
+
+public class UnboundValueVariable extends UnboundVariable implements
+        ValueVariableBuilder,
+        PredicateBuilder<ValueVariable>,
+        ExpressionBuilder<ValueConstraint.Assignment.Expression.ValueVar> {
```
flyingsilverfin, line 37: These should be `PredicateArgument` and `ExpressionArgument` @flyingsilverfin

java/pattern/constraint/ValueConstraint.java
```diff
@@ -0,0 +1,658 @@
...
+
+            public Operation asOperation() {
+                throw TypeQLException.of(INVALID_CASTING.message(className(this.getClass()), className(Operation.class)));
+            }
+
+            public Function asFunction() {
+                throw TypeQLException.of(INVALID_CASTING.message(className(this.getClass()), className(Function.class)));
+            }
+
+            public Bracketed asBracketed() {
```
flyingsilverfin, line 204: Rename to 'Parenthesis' @flyingsilverfin

java/query/builder/Sortable.java
```diff
@@ -41,32 +42,33 @@
 
     default S sort(String var, String... vars) {
```
flyingsilverfin, line 43: Remove String interface since it's not accurate

flyingsilverfin: @flyingsilverfin

PR #266, opened 2023-02-17
Unique attribute ownership

flyingsilverfin:
**For the title of this PR:** please follow the grammatical rules of a usual publication title, without capitalisation (except for the first letter). Thus, the title should NOT CONTAIN CODE: no dots, no parentheses, no backticks, no brackets, etc. It needs to be distinctive (not detailed) and succinct (not lengthy). Details of this PR will go in the description. **For the description of this PR:** please replace every line in curly brackets ( { like this } ) with an appropriate description following the guidance. Finally, **please remove this paragraph**.

{ In the form of a paragraph (only use bullet points if strictly necessary), please describe the goal of this PR, why they are valuable to achieve, and reference the related GitHub issues. This section will be automatically compiled into the release notes, so please:
  - describe the impact of the change in this PR to the _user_ of this repository (e.g. end user, contributor, developer).
  - describe the new product behaviour in _present tense_, and the old behaviour and how it's been changed in _past tense_.
  - Use the _Royal We_: _"We"_ made changes, not _"I"_ made changes. }

{ Please explain what you implemented, why your changes are the best way to achieve the goal(s) above. Please describe every method, class and package, by explaining:
  - its responsibility, 
  - how it's expected to behave, and 
  - how it relates to the adjacent methods/classes/packages it interacts with. 

This would allow the reviewer to understand your intentions in the code much better. If you're adding new classes, make sure these explanations are also included in the class header comments. Last but not least, please reference the GitHub issues to be automatically closed, such like 'closes #number'. }

### DIFF
--- a/grammar/TypeQL.g4
+++ b/grammar/TypeQL.g4
@@ -113,19 +113,15 @@ variable_concept      :   VAR_  IS  VAR_  ;
 variable_type         :   type_any    type_constraint ( ',' type_constraint )*  ;
 type_constraint       :   ABSTRACT
                       |   SUB_        type_any
-                      |   OWNS        type         ( AS type )?     owns_annotations //value_annotations
-                      |   RELATES     type         ( AS type )?   //relates_annotations
+                      |   OWNS        type         ( AS type )?     annotations_owns
+                      |   RELATES     type         ( AS type )?
                       |   PLAYS       type_scoped  ( AS type )?
                       |   VALUE       value_type
                       |   REGEX       STRING_
                       |   TYPE        label_any
-                      // value_annotations
                       ;
-//annotations           :   ( ANNOTATION_KEY )?   ( ANNOTATION_UNIQUE )?  ( ANNOTATION_RANGE | ANNOTATION_REGEX )?     ;
 
-owns_annotations      :   ( ANNOTATION_KEY )?   ( ANNOTATION_UNIQUE )?          ;
-//relates_annotations :
-//value_annotations   :   ( ANNOTATION_RANGE | ANNOTATION_REGEX )?
+annotations_owns      :   ( ANNOTATION_KEY )?   ( ANNOTATION_UNIQUE )?          ;
 
 // THING VARIABLES =============================================================
 
--- a/grammar/TypeQL.g4
+++ b/grammar/TypeQL.g4
@@ -113,13 +113,19 @@ variable_concept      :   VAR_  IS  VAR_  ;
 variable_type         :   type_any    type_constraint ( ',' type_constraint )*  ;
 type_constraint       :   ABSTRACT
                       |   SUB_        type_any
-                      |   OWNS        type         ( AS type )? ( IS_KEY )?
-                      |   RELATES     type         ( AS type )?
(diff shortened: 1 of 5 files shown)


grammar/TypeQL.g4
```diff
@@ -113,13 +113,19 @@ variable_concept      :   VAR_  IS  VAR_  ;
...
+                      |   RELATES     type         ( AS type )?   //relates_annotations
                       |   PLAYS       type_scoped  ( AS type )?
                       |   VALUE       value_type
                       |   REGEX       STRING_
                       |   TYPE        label_any
+                      // value_annotations
                       ;
+//annotations           :   ( ANNOTATION_KEY )?   ( ANNOTATION_UNIQUE )?  ( ANNOTATION_RANGE | ANNOTATION_REGEX )?     ;
+
+owns_annotations      :   ( ANNOTATION_KEY )?   ( ANNOTATION_UNIQUE )?          ;
```
flyingsilverfin, line 126: @haikalpribadi do you think we should use a generic `annotations` block that can be any annotations, and then do the validation once we construct the TypeQL objects (see the commented `annotations` above, which we would apply using:
```as OWNS type (as type)? annotations```

In the future we will have long list of annotations (range, cardinality, regex, etc.) that are possible, but only some will be applicable to `owns`, some will be applicable to `plays` only, some will be applicable to the variable directly... given this I think we should have `owns_annotations` separate, then add over time.

Comment (haikalpribadi) [reply grammar/TypeQL.g4:126]: What if we can provide more than one annotation? 🤔 

Also, I think the it should be `annotation_own` and `annotation_relation`, no?

Comment (flyingsilverfin) [reply grammar/TypeQL.g4:126]: Yes my bad re. annotation_owns, was prototyping too quickly! I think it would be on the constraint eg. `annotation_owns` and `annotation_relates` rather than `annotation_own` and `annotation_relation` though.

This mechanism does allow us to provide more than one annotation. The alternative is inline:
```
OWNS type ( AS type )? ( ANNOTATION_KEY )? ( ANNOTATION_UNIQUE )?
```
which also works but will overflow the line and isn't as nice to read.

haikalpribadi: Why the overflow? You mean the line becomes too long horizontally?

flyingsilverfin: yes

haikalpribadi: why not wrap it in another rule? like what you had before but better named?

flyingsilverfin: you mean the entire row for `type_owns_constraint`, or just keep the `OWNS type` as is and add a new `annotations_owns = ANNOTATION_KEY )? ( ANNOTATION_UNIQUE )?` similar to what I suggested?

krishnangovindraj: `(annotations_owns = ANNOTATION_KEY )? ( ANNOTATION_UNIQUE )?` enforces an order on the annotations
`@key @unique` parses, whereas `@unique @key` does not.

flyingsilverfin:
Replaced by #273

PR #273, opened 2023-03-29
Introduce unique annotation

krishnangovindraj:

We generalise the annotation syntax and parsing to be able to handle a new type of annotation: the `@unique` annotation, which is available only on the `owns` constraint:

```
define
person sub entity, owns email @unique;
email sub attribute, value string;
```

This annotation indicates that any `email`s a person owns must be unique to that person. It does not place any restrictions on the number of emails any given person may own.

The language builder API has also been updated to use a generalised form of any number of annotations, rather than having a particular boolean per annotation type (now, pass annotation `UNIQUE` or `KEY` instead of booleans).

* Introduce `annotations_owns` in the grammar
* Introduce new grammar rules for `annotations`: `@key` becomes an `annotation`, as does the new `@unique`
* Introduce generalised `annotations` in the grammar and the language builders

### DIFF
--- a/grammar/TypeQL.g4
+++ b/grammar/TypeQL.g4
@@ -113,19 +113,15 @@ variable_concept      :   VAR_  IS  VAR_  ;
 variable_type         :   type_any    type_constraint ( ',' type_constraint )*  ;
 type_constraint       :   ABSTRACT
                       |   SUB_        type_any
-                      |   OWNS        type         ( AS type )?     owns_annotations //value_annotations
-                      |   RELATES     type         ( AS type )?   //relates_annotations
+                      |   OWNS        type         ( AS type )?     annotations_owns
+                      |   RELATES     type         ( AS type )?
                       |   PLAYS       type_scoped  ( AS type )?
                       |   VALUE       value_type
                       |   REGEX       STRING_
                       |   TYPE        label_any
-                      // value_annotations
                       ;
-//annotations           :   ( ANNOTATION_KEY )?   ( ANNOTATION_UNIQUE )?  ( ANNOTATION_RANGE | ANNOTATION_REGEX )?     ;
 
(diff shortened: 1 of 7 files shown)


grammar/TypeQL.g4
```diff
@@ -113,14 +113,16 @@ variable_concept      :   VAR_  IS  VAR_  ;
...
-                      |   OWNS        type         ( AS type )? ( IS_KEY )?
+                      |   OWNS        type         ( AS type )?     annotations_owns
                       |   RELATES     type         ( AS type )?
                       |   PLAYS       type_scoped  ( AS type )?
                       |   VALUE       value_type
                       |   REGEX       STRING_
                       |   TYPE        label_any
                       ;
 
+annotations_owns      :   ( ANNOTATION_KEY )?   ( ANNOTATION_UNIQUE )?          ;
```
flyingsilverfin, line 124: @krishnangovindraj  raised good point about order indendence:

```
OWNS type ( AS type )? (annotation_owns*)

annotation_owns = ANNOTATION_KEY | ANNOTATION_UNIQUE | ANNOTATION_CARD ...
```

note that now each one is optional independently and the rule is written in singular form.

java/pattern/constraint/TypeConstraint.java
```diff
@@ -399,46 +408,85 @@ public int hashCode() {
 
     public static class Owns extends TypeConstraint {
 
+        private static final Set<Annotation> VALID_ANNOTATIONS = set(Annotation.KEY, Annotation.UNIQUE);
+
         private final TypeVariable attributeType;
         private final TypeVariable overriddenAttributeType;
-        private final boolean isKey;
+        private final List<Annotation> annotations;
+        private final Annotation uniqueness;
```
flyingsilverfin, line 416: We don't want to group annotations into super-types of annotations in the language for validation/restriction purposes.

PR #278, opened 2023-04-27
Fix TypeQL Python build

dmitrii-ubskii:

We fix the issue with grammar-python producing empty pip packages.

`rules_antlr` cheats: since the `py_library` expects `*.py` sources, `antlr(language = "Python3")` names the package _folder_ `*.py` so that the `py_library()` rule would accept its output as an input. Unfortunately, the current version of `py_library()` is not so easily tricked and produces `error: can't copy 'bazel-bin/grammar/python/python-src.py': doesn't exist or not a regular file`.

In this PR, we explicitly redeclare the python sources produced by the `antlr()` rule for use by `py_library()`.

Drive-by: fix top-level checkstyle to _exclude_ rather than include the bazel remote cache files.

### DIFF
--- a/dependencies/vaticle/repositories.bzl
+++ b/dependencies/vaticle/repositories.bzl
@@ -20,8 +20,8 @@ load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 def vaticle_dependencies():
     git_repository(
         name = "vaticle_dependencies",
-        remote = "https://github.com/vaticle/dependencies",
-        commit = "b968cab9c4d2f85d80a0069b8d15cce5afdd0da5", # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_dependencies
+        remote = "https://github.com/dmitrii-ubskii/vaticle-dependencies",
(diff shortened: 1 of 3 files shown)


dependencies/vaticle/repositories.bzl
```diff
@@ -20,8 +20,8 @@ load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 def vaticle_dependencies():
     git_repository(
         name = "vaticle_dependencies",
-        remote = "https://github.com/vaticle/dependencies",
-        commit = "b968cab9c4d2f85d80a0069b8d15cce5afdd0da5", # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_dependencies
+        remote = "https://github.com/dmitrii-ubskii/vaticle-dependencies",
+        commit = "0164a8ee178c6e54a7b4d624ecb2cdd3cc8389ed", # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_dependencies
```
dmitrii-ubskii, line 24: Pending https://github.com/vaticle/dependencies/pull/439

flyingsilverfin (approved):
Thank you!!

PR #280, opened 2023-05-05
Update release notes workflow

dmitrii-ubskii:

We integrate the new release notes tooling. The release notes are now to be written by a person and committed to the repo.

`tool/release/create_notes.sh` is a manual replacement for automatic `@vaticle_dependencies//tool/release/notes:create` in CI. The script is meant to be executed manually by the writer of the release notes, who will edit the autocompiled notes before committing LATEST_RELEASE_NOTES.md to the repository.

In the Release job in Factory, we check to see whether there were any changes committed since the release notes have last been changed, as a safeguard against accidentally releasing with stale release notes.


tools/release/create_notes.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
```
lolski, line 1: The package should be called `tool` (singular) - we prefer singular to plural when naming packages

dmitrii-ubskii: This folder is more akin to "dependencies" and "tests" than it is to "util" or "test". We have a mild precedent with "resources" in Studio and "templates" in Console. I'm inclined to stick with "tools".

flyingsilverfin: You should stick with `tool` since the the package you're pulling from is called `//.../tool`. Let's be consistent for now and revisit plurasation discussion at another date.

flyingsilverfin: Separate note: only python and rust call it `tests`, while Node and Java use `test`

flyingsilverfin: @dmitrii-ubskii once you rename this I can merge it

flyingsilverfin (approved):

PR #282, opened 2023-05-10
Implement unique annotation in TypeQL Rust

dmitrii-ubskii:

We generalise the annotation syntax and parsing to be able to handle a new type of annotation: the `@unique` annotation, which is available only on the owns constraint: `define person sub entity, owns email @unique;`

The `@unique` annotation has been introduced first in TypeQL Java in https://github.com/vaticle/typeql/pull/273.

We update the pest grammar used in the Rust parser to mirror the changes in TypeQL Java, and adjust the parser visitors to match.

A few drive-by fixes:
- mild refactoring prompted by Clippy;
- replaced usage of deprecated `chrono` `NaiveDateTime` builders with up-to-date variants.

### DIFF
--- a/rust/parser/mod.rs
+++ b/rust/parser/mod.rs
@@ -39,17 +39,15 @@ use crate::{
         Result,
     },
     pattern::{
-        Annotation, ConceptVariable, ConceptVariableBuilder, Conjunction, Definable, Disjunction,
-        HasConstraint, IsaConstraint, Label, Negation, OwnsConstraint, Pattern, PlaysConstraint,
-        RelatesConstraint, RelationConstraint, RolePlayerConstraint, RuleDeclaration,
-        RuleDefinition, SubConstraint, ThingConstrainable, ThingVariable, ThingVariableBuilder,
-        Type, TypeConstrainable, TypeVariable, TypeVariableBuilder, UnboundVariable, Value,
-        ValueConstraint, Variable,
+        Annotation, ConceptVariable, ConceptVariableBuilder, Conjunction, Definable, Disjunction, HasConstraint,
+        IsaConstraint, Label, Negation, OwnsConstraint, Pattern, PlaysConstraint, RelatesConstraint,
+        RelationConstraint, RolePlayerConstraint, RuleDeclaration, RuleDefinition, SubConstraint, ThingConstrainable,
+        ThingVariable, ThingVariableBuilder, Type, TypeConstrainable, TypeVariable, TypeVariableBuilder,
+        UnboundVariable, Value, ValueConstraint, Variable,
     },
     query::{
-        sorting, AggregateQueryBuilder, Query, Sorting, TypeQLDefine, TypeQLDelete, TypeQLInsert,
-        TypeQLMatch, TypeQLMatchAggregate, TypeQLMatchGroup, TypeQLMatchGroupAggregate,
-        TypeQLUndefine, TypeQLUpdate,
+        sorting, AggregateQueryBuilder, Query, Sorting, TypeQLDefine, TypeQLDelete, TypeQLInsert, TypeQLMatch,
+        TypeQLMatchAggregate, TypeQLMatchGroup, TypeQLMatchGroupAggregate, TypeQLUndefine, TypeQLUpdate,
     },
 };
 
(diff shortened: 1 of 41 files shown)


rust/parser/test/mod.rs
```diff
@@ -271,7 +271,10 @@ $x has release-date < 1986-03-03T00:00,
     let expected = typeql_match!(var("x")
         .has((
             "release-date",
-            lt(NaiveDateTime::new(NaiveDate::from_ymd(1986, 3, 3), NaiveTime::from_hms(0, 0, 0)))
+            lt(NaiveDateTime::new(
+                NaiveDate::from_ymd_opt(1986, 3, 3).unwrap(),
+                NaiveTime::from_hms_opt(0, 0, 0).unwrap()
+            ))
```
dmitrii-ubskii, line 277: `from_ymd`, `from_hms`, and `from_hms_millis` are all deprecated, and their new `_opt` variants (which validate the input and return an `Option<T>`) are to be used instead.

rust/parser/test/mod.rs
```diff
@@ -1189,7 +1203,7 @@ $x owns name @key;
 get $x;"#;
 
     let parsed = parse_query(query).unwrap().into_match();
-    let expected = typeql_match!(var("x").owns(("name", KEY))).get(["x"]);
+    let expected = typeql_match!(var("x").owns(("name", Key))).get(["x"]);
```
dmitrii-ubskii, line 1206: `KEY` the constant (which used to stand for `IsKey::Yes`) is replaced by `[Annotation::]Key` the enum variant.

rust/pattern/constraint/thing/value.rs
```diff
@@ -108,7 +108,7 @@ impl Validatable for Value {
         match &self {
             Self::DateTime(date_time) => {
                 if date_time.nanosecond() % 1000000 > 0 {
-                    Err(TypeQLError::InvalidConstraintDatetimePrecision(date_time.clone()))?
+                    Err(TypeQLError::InvalidConstraintDatetimePrecision(*date_time))?
```
dmitrii-ubskii, line 111: `NaiveDateTime` is plain old data that implements `Copy`, i.e. it is never moved.

rust/pattern/schema/rule.rs
```diff
@@ -25,7 +25,6 @@ use std::{fmt, iter};
 use crate::{
     common::{
         error::{collect_err, TypeQLError},
-        string::indent,
```
dmitrii-ubskii, line 28: Became unused after we changed how rules are formatted.

rust/tests/behaviour/main.rs
```diff
@@ -38,7 +38,7 @@ fn main() {
 }
 
 fn parse_query_in_step(step: &Step) -> Query {
-    parse_query(&step.docstring.as_ref().unwrap()).unwrap()
+    parse_query(step.docstring.as_ref().unwrap()).unwrap()
```
dmitrii-ubskii, line 41: Remove unnecessary double referencing.

rust/pattern/constraint/type_/owns.rs
```diff
@@ -20,54 +20,44 @@
...
-        Ok(())
     }
 }
 
 #[derive(Debug, Clone, Eq, PartialEq)]
 pub struct OwnsConstraint {
     pub attribute_type: TypeVariable,
     pub overridden_attribute_type: Option<TypeVariable>,
-    pub is_key: IsKeyAttribute,
+    pub annotations: HashSet<Annotation>,
```
dmitrii-ubskii, line 51: Potentially could be a bitset instead, which may or may not save a bit of overhead.

rust/pattern/constraint/type_/owns.rs
```diff
@@ -20,54 +20,44 @@
...
 }
 
 impl OwnsConstraint {
-    fn new(
+    pub(crate) fn new(
```
dmitrii-ubskii, line 55: `OwnsConstraint` is now created directly by the parser without relying on `from()`. `From` impls now don't have to accept `Option<T>` for the overridden attribute label: it's either present or it isn't.

rust/pattern/constraint/type_/owns.rs
```diff
@@ -89,113 +79,202 @@ impl Validatable for OwnsConstraint {
 
 impl From<&str> for OwnsConstraint {
     fn from(attribute_type: &str) -> Self {
-        OwnsConstraint::from((attribute_type, IsKeyAttribute::No))
+        OwnsConstraint::from(Label::from(attribute_type))
     }
 }
```
dmitrii-ubskii, line 84: A little more systematic in how the `From<T>` are implemented for the constraint:

- `&str`/`String` are transformed into `Label`;
- `Label`, `Type`, and `UnboundVariable` are all transformed into `TypeVariable`;
- single `Annotation` — into an array of `Annotation`s.

dmitrii-ubskii: These could potentially be generated by a clever macro based on the rules like the above, to simplify the cognitive load.

rust/parser/mod.rs
```diff
@@ -483,19 +485,20 @@ fn visit_variable_type(tree: SyntaxTree) -> TypeVariable {
...
                 Rule::OWNS => {
-                    let type_ = visit_type(constraint.consume_expected(Rule::type_));
-                    let overridden = constraint
-                        .consume_if_matches(Rule::AS)
-                        .and_then(|_| Some(visit_type(constraint.consume_expected(Rule::type_))));
-                    let is_key =
-                        IsKeyAttribute::from(constraint.consume_if_matches(Rule::IS_KEY).is_some());
-                    var_type.constrain_owns(OwnsConstraint::from((type_, overridden, is_key)))
+                    let type_ =
+                        visit_type(constraint.consume_expected(Rule::type_)).into_type_variable();
```
flyingsilverfin, line 489: this line looks too short / awkwardly split up. I think it would look  better like:
```
let type = visit_type(...)
    .into_type_variable()
```

rust/pattern/constraint/type_/owns.rs
```diff
@@ -89,113 +79,314 @@ impl Validatable for OwnsConstraint {
...
+impl From<Label> for OwnsConstraint {
+    fn from(attribute_type: Label) -> Self {
+        OwnsConstraint::from(UnboundVariable::hidden().type_(attribute_type))
     }
 }
 
-impl From<(Label, IsKeyAttribute)> for OwnsConstraint {
-    fn from((attribute_type, is_key): (Label, IsKeyAttribute)) -> Self {
-        OwnsConstraint::from((UnboundVariable::hidden().type_(attribute_type), is_key))
+impl From<Type> for OwnsConstraint {
```
flyingsilverfin, line 98: After investigation we found that `Label.Type` shouldn't exist outside of the Parser implementation, which will also require that the constructors here for `Type` will disappear.

rust/pattern/constraint/type_/owns.rs
flyingsilverfin, line 51: Ths should be a list/vector, since we should store the user's query faithfully

flyingsilverfin (changes requested):

flyingsilverfin (approved):

PR #285, opened 2023-05-23
Fix rule validation

dmikhalin:

While parsing a rule with `has` and without relation in `then` part `expect_valid_inference()` function panicked. Now it returns an `Error`.

We incorporated `expect_infer_single_edge()` function into `expect_valid_inference()`.

### DIFF
--- a/rust/pattern/schema/rule.rs
+++ b/rust/pattern/schema/rule.rs
@@ -128,28 +128,26 @@ fn contains_negations<'a>(mut patterns: impl Iterator<Item = &'a Pattern>) -> bo
 fn expect_valid_inference(then: &ThingVariable, rule_label: &Label) -> Result<()> {
     if then.has.len() == 1
         && (then.iid.is_none() && then.isa.is_none() && then.value.is_none() && then.relation.is_none())
-        || then.relation.is_some()
-            && then.isa.is_some()
-            && (then.iid.is_none() && then.has.is_empty() && then.value.is_none())
     {
-        if let Some(has) = then.has.get(0) {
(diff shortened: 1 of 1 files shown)


rust/pattern/schema/rule.rs
```diff
@@ -126,37 +125,33 @@ fn contains_negations<'a>(mut patterns: impl Iterator<Item = &'a Pattern>) -> bo
...
 }
 
-fn expect_infer_single_edge(then: &ThingVariable, rule_label: &Label) -> Result<()> {
+fn expect_valid_inference(then: &ThingVariable, rule_label: &Label) -> Result<()> {
     if then.has.len() == 1
         && (then.iid.is_none() && then.isa.is_none() && then.value.is_none() && then.relation.is_none())
         || then.relation.is_some()
             && then.isa.is_some()
             && (then.iid.is_none() && then.has.is_empty() && then.value.is_none())
     {
```
dmitrii-ubskii, line 134: I'd suggest splitting this condition into two branches, since each side of the `||` corresponds to an inner if-else branch. That should let us get rid of the `unreachable!()` altogether.

dmitrii-ubskii (changes requested):

rust/pattern/schema/rule.rs
dmitrii-ubskii: It will also read better if you factor the conditions out into their own functions, come to think.

dmitrii-ubskii (approved):
LGTM!

flyingsilverfin (approved):

PR #286, opened 2023-05-24
Debug formatting Errors

nikolaimerritt:
We improve the debug formatting for macro-generated errors, by including more detail about the error. This solves issue https://github.com/vaticle/typedb-client-rust/issues/44

We include the error code and error message in the debug format for macro-generated errors. We do this by explicitly implimenting `std::fmt::Debug`, inserting the error code and error message.

### DIFF
--- a/rust/common/error/macros.rs
+++ b/rust/common/error/macros.rs
@@ -63,9 +63,9 @@ macro_rules! error_messages {
                 }
             }
 
-            fn name(&self) -> String {
+            const fn name(&self) -> &'static str {
                 match self {$(
-                    Self::$error_name(..) => format!("{}::{}", std::any::type_name::<Self>(), std::stringify!($error_name)),
+                    Self::$error_name(..) => concat!(std::stringify!($name), "::", std::stringify!($error_name)),
                 )*}
             }
         }
@@ -84,11 +84,9 @@ macro_rules! error_messages {
 
         impl std::fmt::Debug for $name {
             fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
-                let mut binding = f.debug_struct(&self.name());
-                let mut debug_struct = binding
-                    .field("message", &format!("{}", self));
-                $(error_messages!(@payload self, $error_name, debug_struct $(, $inner)*);)*
-                debug_struct.finish()
+                let mut binding = f.debug_struct(self.name());
+                $(error_messages!(@payload self, $error_name, binding.field("message", &format!("{}", self)) $(, $inner)*);)*
+                binding.finish()
             }
         }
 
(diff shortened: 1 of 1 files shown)


rust/common/error/macros.rs
```diff
@@ -109,3 +124,33 @@ macro_rules! error_messages {
...
+            TestError::ErrorWithAttributes(1, "error message".to_string()),
+            TestError::MultiLine(),
+        ];
+
+        for error in errors {
+            let display = format!("{error}");
+            let compact_debug = format!("{error:?}");
+            let expanded_debug = format!("{error:#?}");
+            assert!(compact_debug.contains(&display.replace('\n', &"\\n")));
+            assert!(expanded_debug.contains(&display.replace('\n', &"\\n")));
```
dmitrii-ubskii, line 153: How is this not failing? `display` includes e.g. `[TST1]` at the start which the debug format omits, as far as I can tell.

nikolaimerritt: The `message` field in `std::Debug` has the entire error message produced by `std::Display`.

rust/common/error/macros.rs
```diff
@@ -76,6 +82,15 @@ macro_rules! error_messages {
             }
         }
 
+        impl std::fmt::Debug for $name {
+            fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+                f.debug_struct(&self.name())
+                    .field("code", &self.code())
+                    .field("message", &format!("{self}"))
```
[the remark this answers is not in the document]
nikolaimerritt: `$code_pfx` is included in the message.

rust/common/error/macros.rs
```diff
@@ -76,6 +82,15 @@ macro_rules! error_messages {
...
+                    .finish()
+            }
+        }
+
```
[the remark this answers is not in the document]
nikolaimerritt: Good spot. We now include the error payload, as a field that is a (possibly empty) tuple.

rust/common/error/macros.rs
```diff
@@ -76,6 +89,16 @@ macro_rules! error_messages {
             }
         }
 
+        impl std::fmt::Debug for $name {
+            fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+                let mut debug_struct = f.debug_struct(&self.name());
```
dmitrii-ubskii, line 94: This would call the struct e.g. `MultiLine` instead of the more appropriate `TestError::Multiline`

rust/common/error/macros.rs
```diff
@@ -62,6 +62,19 @@ macro_rules! error_messages {
...
+                    Self::$error_name(..) => std::stringify!($error_name),
+                )*}
+            }
+
+            fn format_payload<'a, 'b>(&'a self, debug_struct: &'a mut std::fmt::DebugStruct<'a, 'b>) -> &mut std::fmt::DebugStruct<'a, 'b>
+                where 'b: 'a
+            {
+                $(error_messages!(@payload self, debug_struct, $error_name $(, $inner)*);)*
+                unreachable!()
+            }
```
dmitrii-ubskii, line 77: Does this helper need to exist? This is specific to `fmt::Debug` and could be inlined.

rust/common/error/macros.rs
```diff
@@ -108,4 +131,65 @@ macro_rules! error_messages {
             return format!($body, _0, _1, _2, _3)
         }
     };
+
+    (@payload $self:ident, $format:ident, $error_name:ident) => {
```
dmitrii-ubskii, line 135: It'd make more sense for `$error_name` to come before `$format` here, as it belongs with `$self`.

rust/common/error/macros.rs
```diff
@@ -108,4 +131,65 @@ macro_rules! error_messages {
...
+        if let Self::$error_name() = &$self {
+            return $format.field("payload", &())
+        }
+    };
```
dmitrii-ubskii, line 139: I wonder if this can be merged with `@format` into a more generic helper that is invoked as e.g. 
```
error_messages!(@dispatch self, $error_name $(, $inner)*, debug_struct.field("payload", &payload));)*
```

dmitrii-ubskii: Might be feasible using an immediately invoked lambda in the implementation.

dmitrii-ubskii (changes requested):

rust/common/error/macros.rs
```diff
@@ -76,6 +89,16 @@ macro_rules! error_messages {
             }
         }
 
+        impl std::fmt::Debug for $name {
+            fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+                let mut debug_struct = f.debug_struct(&self.name());
+                let mut repr = debug_struct
+                    .field("message", &format!("{}", self));
+                repr = self.format_payload(repr);
+                repr.finish()
```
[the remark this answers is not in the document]
nikolaimerritt: `debug_struct` is a binding to the `DebugStruct`. We create this binding to be able to pass the `DebugStruct` to the `@payload` macro without the `DebugStruct` being dropped.

rust/common/error/macros.rs
nikolaimerritt: It doesn't need to exist, and has been removed.

rust/common/error/macros.rs
nikolaimerritt: We now use the full type name, as per https://doc.rust-lang.org/std/any/fn.type_name.html

rust/common/error/macros.rs
nikolaimerritt: Now that writing the payload field has been inlined, `@payload` and `@format` seem too different to be reconciled. `@format` returns a value, whereas `@payload` now modifies the `&mut DebugStruct` passed to it.

dmitrii-ubskii (changes requested):

rust/common/error/macros.rs
nikolaimerritt: We've agreed in-person that there is no easy way to unify `@format` with `@payload`. The fundamental issue was in dealing with the unknown number of values wrapped by the enum instance.

PR #294, opened 2023-07-07
Introduce expressions and computed value variables in typeql-rust

dmikhalin:

We introduce in `typeql-rust` the ability to perform arithmetic computation and store the results in a "value variable" - denoted by a preceding `?`, the same as introduced in `typeql-java` by https://github.com/vaticle/typeql/pull/260.

All redundant parenthesis from original query will not persist in the string representation of the parsed expression anymore. If we get this query: 
```
match
  $p isa person, has salary $s;
  ?net = (($s - 12500) * 0.8 + 12500);
```
it will be transformed into
```
match
  $p isa person, has salary $s;
  ?net = ($s - 12500) * 0.8 + 12500;
```

In addition to changes implemented in https://github.com/vaticle/typeql/pull/260:
- We removed `Parenthesis` class and do not store parenthesis in the expression tree anymore. When we build a string representation of the expression, we add all necessary parenthesis. Nevertheless, if we get something like `a + (b + c)`, these parenthesis will persist in the resulting string representation.


rust/pattern/variable/builder/expression.rs
dmitrii-ubskii, line 1: 1) `Expression` doesn't belong under `builder`. I'd probably put it under `pattern`.
2) Where are the builder functions?

rust/pattern/variable/mod.rs
```diff
@@ -50,25 +57,49 @@ pub enum Variable {
     Concept(ConceptVariable),
     Thing(ThingVariable),
     Type(TypeVariable),
-    Unbound(UnboundVariable),
+    Value(ValueVariable),
+    Unbound(UnboundConceptVariable),
```
dmitrii-ubskii, line 61: Do we really need this rename then if an `Unbound` variable is only ever a `Concept`?

rust/pattern/variable/thing.rs
```diff
@@ -37,13 +37,13 @@ pub struct ThingVariable {
...
     pub has: Vec<HasConstraint>,
-    pub value: Option<ValueConstraint>,
+    pub value: Option<Predicate>,
     pub relation: Option<RelationConstraint>,
 }
 
 impl ThingVariable {
     pub fn new(reference: Reference) -> ThingVariable {
-        ThingVariable { reference, iid: None, isa: None, has: Vec::new(), value: None, relation: None }
+        ThingVariable { reference, iid: None, isa: None, has: Vec::new(), value: None::<Predicate>, relation: None }
```
dmitrii-ubskii, line 46: Redundant turbofish.

rust/pattern/variable/unbound_value.rs
dmitrii-ubskii, line 1: Can we ditch this and always use `ValueVariable`?

dmitrii-ubskii (changes requested):

rust/tests/behaviour/main.rs
```diff
@@ -20,21 +20,24 @@
  *
  */
 
-use cucumber::{gherkin::Step, given, then, when, World};
+#![recursion_limit = "256"]
+
```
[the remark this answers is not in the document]
dmikhalin: Without `recursion_limit` increasing we get "recursion limit reached while expanding `#[when]`".

rust/common/token.rs
```diff
@@ -162,3 +164,21 @@ string_enum! { Order
     Asc = "asc",
     Desc = "desc",
 }
+
+string_enum! { Operation
```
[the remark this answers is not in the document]
dmikhalin: We already have one `Operator` for logic operators. `LogicOperator` and `ArithmeticOperator`?

rust/parser/mod.rs
```diff
@@ -642,6 +690,63 @@ fn visit_predicate(tree: SyntaxTree) -> ValueConstraint {
...
+            Expression::Operation(Operation { op, left: Box::new(left), right: Box::new(right) })
+        })
+        .parse(tree.into_children())
+}
+
+fn visit_function(tree: SyntaxTree) -> Function {
+    let mut children = tree.into_children();
+    let symbol = visit_function_name(children.consume_expected(Rule::expression_function_name));
+    let args = children.map(visit_function_argument).collect();
+    Function { symbol, args }
```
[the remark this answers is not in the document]
dmikhalin: `function_name` looks better to me too. Just copied `symbol` from java.
We have `min` and `max` with multiple args.

rust/pattern/variable/mod.rs
dmikhalin: Discussed in person.

rust/pattern/variable/unbound_value.rs
dmikhalin: Discussed in person.

rust/pattern/constraint/predicate.rs
```diff
@@ -91,7 +97,8 @@ pub enum Value {
     Boolean(bool),
     String(String),
     DateTime(NaiveDateTime),
-    Variable(Box<ThingVariable>),
+    ThingVariable(Box<ThingVariable>),
+    ValueVariable(Box<ValueVariable>),
```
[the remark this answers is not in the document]
dmikhalin: `enum Variable` is too common, there are other variants too. Do you mean we can store only a `Reference`?

rust/builder/mod.rs
```diff
@@ -87,42 +87,50 @@ pub fn var(var: impl Into<UnboundVariable>) -> UnboundVariable {
     var.into()
 }
```
dmitrii-ubskii, line 88: Is this ever valid now?

rust/common/error/mod.rs
```diff
@@ -97,6 +97,8 @@ error_messages! { TypeQLError
         6: "The query has not been provided with any definables.",
     MatchHasNoBoundingNamedVariable() =
         7: "The match query does not have named variables to bound the nested disjunction/negation pattern(s).",
+    VariableNameConflict(String) =
+        8: "The variable(s) named '{}' cannot be used for both concept variables and value variables.",
```
dmitrii-ubskii, line 101:
```suggestion
    VariableNameConflict(String) =
        8: "The variable name '{}' cannot be used for both concept variables and value variables.",
```
It's the name that can't be reused, not the variables.

rust/builder/mod.rs
```diff
@@ -21,10 +21,10 @@
  */
 
 use crate::{
-    common::token::Predicate,
+    common::token,
```
dmitrii-ubskii, line 24: Can we revert this now?

rust/tests/behaviour/main.rs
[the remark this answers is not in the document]
dmitrii-ubskii: Following in-person discussion: the culprit is the 43 `#[step(...)]` attributes, which get expanded into 129 cucumber proc-macros — just over the default limit of 128. Meaningfully splitting the two sets of no-op steps was deemed preferable to raising the recursion limit.

rust/common/token.rs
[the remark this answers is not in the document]
dmitrii-ubskii: `LogicalOperator`, but yes, I think this is better.

rust/pattern/constraint/predicate.rs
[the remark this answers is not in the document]
dmitrii-ubskii: No. This is fine, stronger typing is preferable.

rust/parser/test/mod.rs
```diff
@@ -427,7 +416,144 @@ $x isa movie,
...
+}
+
+#[test]
+fn test_attribute_query_by_value_variable() {
+    let query = r#"match
+?x = 5;
+$a == ?x isa age;"#;
+
+    let parsed = parse_query(query).unwrap().into_match();
+    let expected = typeql_match!(vvar("x").assign(Expression::Constant(5.into())), cvar("a").eq(vvar("x")).isa("age"),);
```
dmitrii-ubskii, line 431:
```suggestion
    let expected = typeql_match!(vvar("x").assign(5), cvar("a").eq(vvar("x")).isa("age"));
```

rust/parser/test/mod.rs
```diff
@@ -427,7 +416,144 @@ $x isa movie,
...
+(commodity: $x, qty: $q) isa order;
+?net = $p * $q;
+?gross = ?net * ( 1.0 + 0.21 );"#;
+
+    let parsed = parse_query(query).unwrap().into_match();
+    let expected = typeql_match!(
+        cvar("x").isa("commodity").has(("price", cvar("p"))),
+        rel(("commodity", "x")).rel(("qty", "q")).isa("order"),
+        vvar("net").assign(cvar("p").multiply(cvar("q"))),
+        vvar("gross").assign(vvar("net").multiply(1.0.add(0.21))),
```
dmitrii-ubskii, line 460: I've got mixed feelings about `1.0.add(0.21)`
One one hand, this is neat. On the other, this kind of clashes with `f64 as Add<f64>` [1]. I'm not prepared to say that this is a problem: if the user writes `1.add(2)` in the builder and TypeQL produces either `3` or `1+2`, those should be indistinguishable to the user. I'm not sure the law of least surprise kicks in at any point here, but that remains to be seen for smth like strings.

[1] <https://doc.rust-lang.org/stable/std/primitive.f64.html#impl-Add%3Cf64%3E-for-f64>

rust/parser/test/mod.rs
```diff
@@ -617,12 +738,7 @@ get $x, $y;
 count;"#;
 
     let parsed = parse_query(query).unwrap().into_aggregate();
-    let expected = typeql_match!(rel("x").rel("y").isa("friendship"))
-        .get([
-            UnboundVariable::Concept(UnboundConceptVariable::named(String::from("x"))),
-            UnboundVariable::Concept(UnboundConceptVariable::named(String::from("y"))),
-        ])
-        .count();
+    let expected = typeql_match!(rel("x").rel("y").isa("friendship")).get([cvar("x").into(), cvar("y").into()]).count();
```
dmitrii-ubskii, line 741: There must be a way for us to write `.get([cvar("x"), cvar("y")])` instead of `.get([cvar("x").into(), cvar("y").into()])`

rust/pattern/expression/operation.rs
```diff
@@ -0,0 +1,177 @@
...
+    pub fn new(op: OperationToken, left: impl Into<Expression>, right: impl Into<Expression>) -> Self {
+        let left = Box::new(left.into());
+        let right = Box::new(right.into());
+        match op {
+            OperationToken::Add => Operation::Addition { left, right },
+            OperationToken::Subtract => Operation::Subtraction { left, right },
+            OperationToken::Multiply => Operation::Multiplication { left, right },
+            OperationToken::Divide => Operation::Division { left, right },
+            OperationToken::Modulo => Operation::Modulo { left, right },
+            OperationToken::Power => Operation::Power { left, right },
```
dmitrii-ubskii, line 48: `OperationToken` => `token::Operation`

dmitrii-ubskii: Or indeed `ArithmeticOperator`

rust/pattern/expression/operation.rs
```diff
@@ -0,0 +1,177 @@
...
+
+impl Precedence {
+    pub(crate) fn level(&self) -> usize {
+        match self {
+            Precedence::Addition => 3,
+            Precedence::Multiplication => 2,
+            Precedence::Exponentiation => 1,
+        }
+    }
+}
```
dmitrii-ubskii, line 165: We should instead define `PartialOrd` and `Ord` for `Precedence` directly.

rust/pattern/expression/operation.rs
```diff
@@ -0,0 +1,177 @@
...
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+enum Precedence {
+    Addition,
+    Multiplication,
+    Exponentiation,
+}
```
dmitrii-ubskii, line 155: This should be `Copy`. This goes for all simple enums like this.

rust/pattern/variable/builder/expression.rs
```diff
@@ -20,107 +20,92 @@
...
+    fn multiply(self, right: impl Into<Expression>) -> Expression;
+    fn divide(self, right: impl Into<Expression>) -> Expression;
+    fn modulo(self, right: impl Into<Expression>) -> Expression;
+    fn power(self, right: impl Into<Expression>) -> Expression;
+    fn abs(arg: impl Into<Expression>) -> Expression;
+    fn ceil(arg: impl Into<Expression>) -> Expression;
+    fn floor(arg: impl Into<Expression>) -> Expression;
+    fn max<const N: usize>(args: [impl Into<Expression>; N]) -> Expression;
+    fn min<const N: usize>(args: [impl Into<Expression>; N]) -> Expression;
+    fn round(arg: impl Into<Expression>) -> Expression;
```
dmitrii-ubskii, line 40: Static functions on a trait here don't make a whole lot of sense. You'll have `Expression::round(1.2)`, but also `UnboundValueVariable::round(1.2)` as valid invocations.

These should be top-level, next to the top-level predicates such as `eq()` and `gte()`.

rust/pattern/variable/builder/expression.rs
```diff
@@ -20,107 +20,92 @@
...
+        Expression::max([$($args, )*])
+    }}
+}
+
+#[macro_export]
+macro_rules! min {
+    ($($args:expr),*) => {{
+        Expression::min([$($args, )*])
+    }}
 }
```
dmitrii-ubskii: These should also move to the top-level `builder` module.

rust/pattern/expression/function.rs
```diff
@@ -0,0 +1,49 @@
...
+    }
+}
+
+impl fmt::Display for Function {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(
+            f,
+            "{}({})",
+            self.function_name,
+            self.args.iter().map(|expr| expr.to_string()).collect::<Vec<_>>().join(", ")
```
dmitrii-ubskii, line 46: Didn't we have a macro for this pattern that doesn't allocate the needless `Vec`?

dmitrii-ubskii (changes requested):

rust/tests/behaviour/main.rs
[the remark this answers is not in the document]
dmikhalin: Done

rust/builder/mod.rs
dmikhalin: We use it in `Sorting`, that can get any `UnboundVariable`.

rust/common/error/mod.rs
dmikhalin: Yes, but plural.

rust/parser/test/mod.rs
dmikhalin: We can create a function `constant` like as `cvar` and `vvar`.

rust/parser/test/mod.rs
dmikhalin: We can create a function `constant` like `cvar` and `vvar` and use `constant(1).add(2)` instead of `1.add(2)`.

rust/parser/test/mod.rs
dmikhalin: Added `filter` macro after the in-person discussion. The reason: we can't have an array with mixed types.

rust/pattern/variable/builder/expression.rs
```diff
@@ -21,91 +21,71 @@
...
 use crate::{
-    common::token::{Function as FunctionToken, Operation as OperationToken},
-    pattern::expression::{Expression, Function, Operation},
+    common::token::ArithmeticOperator,
+    pattern::expression::{Expression, Operation},
 };
 
+pub trait SubExpression {
+    fn into_expression(self) -> Expression;
+}
```
dmitrii-ubskii, line 30: There is a simpler way:
```suggestion
pub trait SubExpression: Into<Expression> {}
```

rust/common/error/mod.rs
```diff
@@ -98,7 +98,7 @@ error_messages! { TypeQLError
     MatchHasNoBoundingNamedVariable() =
         7: "The match query does not have named variables to bound the nested disjunction/negation pattern(s).",
     VariableNameConflict(String) =
-        8: "The variable(s) named '{}' cannot be used for both concept variables and value variables.",
+        8: "The variable names '{}' cannot be used for both concept variables and value variables.",
```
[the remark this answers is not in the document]
dmikhalin: We have a string with all affected variables here.

rust/builder/mod.rs
```diff
@@ -104,33 +130,63 @@ pub fn rel<T: Into<RolePlayerConstraint>>(value: T) -> ThingVariable {
...
+pub fn floor<T: Into<Expression>>(arg: T) -> Expression {
+    Expression::Function(Function { function_name: FunctionToken::Floor, args: vec![Box::from(arg.into())] })
+}
+
+pub fn max<const N: usize, T: Into<Expression>>(args: [T; N]) -> Expression {
+    Expression::Function(Function {
+        function_name: FunctionToken::Max,
+        args: args.into_iter().map(|arg| Box::new(arg.into())).collect(),
+    })
+}
```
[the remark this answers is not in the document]
dmikhalin: Done. But now we need to import `token`, `Expression` and `Function` when we use this macro.

rust/builder/mod.rs
dmikhalin: Removed

rust/pattern/expression/mod.rs
```diff
@@ -30,6 +30,7 @@ use chrono::NaiveDateTime;
 pub use constant::Constant;
 pub use function::Function;
 pub use operation::Operation;
+use pest::pratt_parser::Op;
```
dmitrii-ubskii, line 33: ?

rust/query/typeql_match.rs
```diff
@@ -182,9 +182,10 @@ fn expect_variable_names_are_unique(conjunction: &Conjunction) -> Result<()> {
     let value_names = value_refs.iter().map(|r| r.name()).collect::<HashSet<_>>();
     let common_refs = concept_names.intersection(&value_names).collect::<HashSet<_>>();
     if !common_refs.is_empty() {
-        Err(TypeQLError::VariableNameConflict(common_refs.iter().map(|r| r.to_string()).join(", ")))?
+        Err(TypeQLError::VariableNameConflict(common_refs.iter().map(|r| r.to_string()).join(", ")).into())
+    } else {
+        Ok(())
     }
```
dmitrii-ubskii: Either remove the else and use explicit `return` for `Err`, or flip the condition. I'd go with the former, that sounds more like what we're trying to check for.

dmitrii-ubskii (changes requested):

rust/builder/mod.rs
```diff
@@ -83,46 +120,70 @@ pub fn rule(name: &str) -> RuleDeclaration {
...
-    ValueConstraint::new(Predicate::Contains, Value::from(value.into()))
+pub fn floor<T: Into<Expression>>(arg: T) -> Function {
+    Function { function_name: token::Function::Floor, args: vec![Box::from(arg.into())] }
 }
 
-pub fn like<T: Into<String>>(value: T) -> ValueConstraint {
-    ValueConstraint::new(Predicate::Like, Value::from(value.into()))
+pub fn round<T: Into<Expression>>(arg: T) -> Function {
+    Function { function_name: token::Function::Round, args: vec![Box::from(arg.into())] }
 }
```
dmitrii-ubskii, line 189: These builders aren't actually visible from outside the crate, you need to export them in `typeql_lang.rs`

rust/common/error/macros.rs
```diff
@@ -168,7 +168,6 @@ macro_rules! error_messages {
 
 #[cfg(test)]
 mod tests {
```
dmitrii-ubskii, line 170: Drive-by: this needs `#[allow(dead_code)]` so that rust doesn't generate warnings for the tests-only generated code.

dmitrii-ubskii (approved):

rust/pattern/constraint/predicate.rs
```diff
@@ -91,7 +97,8 @@ pub enum Value {
     Boolean(bool),
     String(String),
     DateTime(NaiveDateTime),
-    Variable(Box<ThingVariable>),
+    ThingVariable(Box<ThingVariable>),
```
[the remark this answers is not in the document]
flyingsilverfin: Further, we should have `expression` be a package directly under `pattern`, and `value` be a very small package under `pattern` with this `Value` struct inside

rust/pattern/expression/constant.rs
```diff
@@ -0,0 +1,44 @@
...
+    pub(crate) value: Value,
+}
+
+impl<T: Into<Value>> From<T> for Constant {
+    fn from(value: T) -> Self {
+        Constant { value: value.into() }
+    }
+}
+
+impl SubExpression for Constant {}
```
flyingsilverfin, line 38: Better naming: `ExpressionBuilder`

rust/pattern/expression/operation.rs
```diff
@@ -0,0 +1,194 @@
...
+
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+enum Precedence {
+    Addition,
+    Multiplication,
+    Exponentiation,
+}
+
+impl Precedence {
+    fn discriminant(&self) -> usize {
```
flyingsilverfin, line 163: can we inline this into the `enum` directly? cc @dmitrii-ubskii

rust/typeql_lang.rs
```diff
@@ -29,7 +29,9 @@ pub mod query;
 #[macro_use]
 mod util;
 
-pub use builder::{contains, eq, gt, gte, like, lt, lte, neq, not, rel, rule, type_, var};
+pub use builder::{
+    abs, ceil, constant, contains, cvar, eq, floor, gt, gte, like, lt, lte, neq, not, rel, round, rule, type_, vvar,
```
flyingsilverfin, line 33: Do we also want `min/max`? Double check we have everything we want here.

How many places do we have to update when we add a new function? We should add a piece of documentation to the PR description to make a note of where to change things if we add new functions/operators

flyingsilverfin (commented):
I've requested some changes, but I still want to finish the review around the architecture & usage of `builder`, `constant` and `value`

rust/pattern/expression/constant.rs
```diff
@@ -0,0 +1,44 @@
...
+ * under the License.
+ *
+ */
+
+use std::fmt;
+
+use crate::pattern::{SubExpression, Value};
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct Constant {
```
flyingsilverfin, line 28: Let's collapse `Constant` and `Value`

rust/parser/test/mod.rs
```diff
@@ -398,7 +417,205 @@ $x isa movie,
...
+    let query = r#"match\n
+$z isa person, has age $y;
+?y = $y;"#;
+
+    let parsed = parse_query(query);
+    assert!(parsed.is_err());
+}
+
+#[test]
+fn test_assign_ops() {
```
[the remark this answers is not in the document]
dmikhalin: We have more specific tests now. This one has been removed.

rust/parser/typeql.pest
```diff
@@ -114,73 +114,104 @@ type_constraint = { ABSTRACT
...
+// EXPRESSION CONSTRUCTS =======================================================
+
+expression = { expression_base ~ ( infix ~ expression_base )* }
+
+expression_base = _{ VAR_
+                   | expression_function | value
+                   | paren_expr
+                   }
+
+infix = _{ POWER | MULTIPLY | DIVIDE | MODULO | ADD | SUBTRACT }
```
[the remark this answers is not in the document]
dmikhalin: - `arithmetic_operator`, because we have this name in `token.rs`.
- `infix_operator` or `binary_operator`, because if we add unary operator(s), they can't be united with binary in this case. `binary` is consistent with existent `Arity` enum, prepared for future usage.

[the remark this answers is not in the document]
benjaminasmall: @flyingsilverfin : "EXPONENTIATE"?

rust/pattern/expression/operation.rs
dmikhalin: If we remove this method, we need to repeat this match twice in `Ord`. And for me it's nice to have explicit levels of precedence.

rust/tests/behaviour/main.rs
```diff
@@ -58,89 +59,95 @@ macro_rules! generic_step_impl {
...
-}
+    #[step("get answer of typeql match aggregate")]
+    #[step("get answers of typeql match group aggregate")]
+    #[step("get answers of typeql match group")]
+    #[step("get answers of typeql match")]
+    #[step("reasoning query")]
+    #[step("verify answer set is equivalent for query")]
+    async fn typeql_match(_: &mut TypeQLWorld, step: &Step) {
+        let parsed = parse_query_in_step(step);
+        assert_eq!(parsed, reparse_query(&parsed));
```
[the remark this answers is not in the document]
dmikhalin: Yep, we don't compare a string representation of parsed query with original query string.

rust/pattern/variable/reference/concept.rs
```diff
@@ -24,29 +24,38 @@ use std::fmt;
...
 #[derive(Debug, Clone, Eq, Hash, PartialEq)]
-pub enum Reference {
+pub enum ConceptReference {
     Anonymous(Visibility),
     Name(String),
 }
 
-impl Reference {
+impl ConceptReference {
+    pub(crate) const ANONYMOUS_NAME: &'static str = "_";
```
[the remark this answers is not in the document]
dmikhalin: It shouldn't be public.

rust/typeql_lang.rs
dmikhalin: `min` and `max` are macros and are imported in another way.
Aren't all places needs changing obvious by searching for all mentions of existing functions?

rust/builder/mod.rs
```diff
@@ -75,6 +76,42 @@ macro_rules! or {
     }}
 }
 
+#[macro_export]
+macro_rules! max {
```
flyingsilverfin, line 80: Please group these the right part of this document

rust/parser/test/mod.rs
```diff
@@ -392,13 +411,235 @@ fn when_parsing_date_error_when_handling_overly_precise_nanos() {
...
+        rel(("commodity", "x")).rel(("qty", "q")).isa("order"),
+        vvar("net").assign(cvar("p").multiply(cvar("q"))),
+        vvar("value").assign(floor(vvar("net").multiply(1.21)))
+    );
+
+    assert_valid_eq_repr!(expected, parsed, query);
+}
+
+#[test]
+fn test_function_round() {
```
flyingsilverfin, line 628: These are nicely named now!

rust/parser/test/mod.rs
```diff
@@ -1102,13 +1363,31 @@ fn test_parsing_label() {
...
 }
 
+#[test]
+fn test_rule_attach_attribute_by_value() {
+    let query = r#"define
+rule attach-val: when {
+    $x has age $a;
+    ?d = $a * 365;
+} then {
+    $x has days == ?d;
```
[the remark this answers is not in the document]
flyingsilverfin: Update: confirm this is possible via a different grammar branch (is a syntactic sugar). We should add a test to make sure it parses.

rust/parser/typeql.pest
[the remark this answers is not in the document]
flyingsilverfin: Discussed in person

rust/parser/typeql.pest
```diff
@@ -114,73 +114,103 @@ type_constraint = { ABSTRACT
...
-schema_native = { RULE }
+// LITERAL INPUT VALUES ========================================================
 
 type_native = { THING | ENTITY | ATTRIBUTE | RELATION | ROLE }
 
 value_type = { LONG | DOUBLE | STRING | BOOLEAN | DATETIME }
-value = { STRING_ | DATETIME_ | DATE_ | DOUBLE_ | LONG_ | BOOLEAN_ }
-regex = { STRING_ }
+constant = { STRING_ | DATETIME_ | DATE_ | BOOLEAN_
+          | signed_double | signed_long
```
flyingsilverfin, line 200: space alignment is off here and below by 1

rust/pattern/constraint/predicate.rs
```diff
@@ -86,88 +92,68 @@ impl fmt::Display for ValueConstraint {
...
-    fn from(bool: bool) -> Self {
-        Value::Boolean(bool)
-    }
-}
-
-impl From<&str> for Value {
-    fn from(string: &str) -> Self {
-        Value::String(String::from(string))
+impl<T: Into<Constant>> From<T> for Value {
+    fn from(literal: T) -> Self {
```
flyingsilverfin, line 112: more `literal` -- do a search over the whole project for "literal" just in case :)

rust/pattern/variable/mod.rs
```diff
@@ -81,7 +113,8 @@ enum_wrapper! { Variable
     ConceptVariable => Concept,
     ThingVariable => Thing,
     TypeVariable => Type,
-    UnboundVariable => Unbound,
+    ValueVariable => Value,
+    UnboundConceptVariable => Unbound,
```
flyingsilverfin, line 117: Not sure what this is but does this need a `UnboundValueVariable` variant?

flyingsilverfin (changes requested):

rust/pattern/expression/operation.rs
dmikhalin: It is a standard practice for attaching constants to `enum` variants.

flyingsilverfin (approved):
Great work!

PR #296, opened 2023-07-12
Relation variable name in rule 'then' must not be present.

shiladitya-mukherjee:

Check that relation variables used to infer new relations in then clause should be anonymous

- Updated TypeDB behaviour dependency.
- Added error message to display in case of exception thrown from the check.
- Added test to check if the relation reference variable in then clause is anonymous.

### DIFF
--- a/java/common/exception/ErrorMessage.java
+++ b/java/common/exception/ErrorMessage.java
@@ -91,8 +91,8 @@ public class ErrorMessage extends com.vaticle.typedb.common.exception.ErrorMessa
             new ErrorMessage(34, "Rule '%s' 'then' variables must be present in the 'when', outside of nested patterns.");
     public static final ErrorMessage INVALID_RULE_THEN_ROLES =
             new ErrorMessage(35, "Rule '%s' 'then' '%s' must specify all role types explicitly or by using a variable.");
-    public static final ErrorMessage INVALID_RULE_THEN_EXPLICIT_VARIABLE =
-            new ErrorMessage(36, "Rule '%s': relation variable '%s' in 'then' must not be explicit.");
+    public static final ErrorMessage INVALID_RULE_THEN_RELATION_VARIABLE =
+            new ErrorMessage(36, "Rule '%s': relation variable name '%s' in 'then' must not be present.");
     public static final ErrorMessage REDUNDANT_NESTED_NEGATION =
             new ErrorMessage(37, "Invalid query containing redundant nested negations.");
     public static final ErrorMessage VARIABLE_NOT_SORTED =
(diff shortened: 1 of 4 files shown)


java/common/exception/ErrorMessage.java
```diff
@@ -105,6 +105,8 @@ public class ErrorMessage extends com.vaticle.typedb.common.exception.ErrorMessa
             new ErrorMessage(41, "'%s' is not a valid Type label. Type labels must start with a letter, and may contain only letters, numbers, '-' and '_'.");
     public static final ErrorMessage INVALID_ANNOTATION =
             new ErrorMessage(42, "Invalid annotation '%s' on '%s' constraint");
+    public static final ErrorMessage RELATION_NOT_ANONYMOUS =
```
krishnangovindraj, line 108: This might fit better close to `INVALID_RULE_THEN_ROLES`.
(This will indeed require you to increase every following key by 1. Unchanging ErrorCodes will hopefully be in TypeDB 3)

Can the name be updated to reflect the fact that it's about  `then` clauses?

java/pattern/schema/Rule.java
```diff
@@ -163,8 +164,16 @@ private static void validateThen(String label, @Nullable Conjunction<? extends P
                 .reduce(true, Boolean::logicalAnd)) {
             throw TypeQLException.of(INVALID_RULE_THEN_ROLES.message(label, then));
         }
+
+        // relation variables used to infer new relations in then clause should be anonymous
+        if (then.relation().isPresent())
+        {
+            if (then.reference().isName())
+                throw TypeQLException.of(RELATION_NOT_ANONYMOUS.message(label, then));
```
krishnangovindraj, line 172: Can combine the conditions with &&. Also, stray line below

krishnangovindraj (changes requested):

java/common/exception/ErrorMessage.java
shiladitya-mukherjee: how about RELATION_IN_THEN_NOT_ANONYMOUS?

java/pattern/schema/Rule.java
```diff
@@ -163,8 +164,14 @@ private static void validateThen(String label, @Nullable Conjunction<? extends P
                 .reduce(true, Boolean::logicalAnd)) {
             throw TypeQLException.of(INVALID_RULE_THEN_ROLES.message(label, then));
         }
+
+        // relation variables used to infer new relations in then clause should be anonymous
+        if (then.relation().isPresent()&&then.reference().isName()) {
+                throw TypeQLException.of(RELATION_IN_THEN_NOT_ANONYMOUS.message(label, then));
```
krishnangovindraj, line 170: Careful. You've copied the call from a message that has two format specifiers ('%s'), but yours doesn't have any.
(Also, please apply auto-spacing with Cmd+L)

java/pattern/schema/Rule.java
```diff
@@ -163,8 +164,14 @@ private static void validateThen(String label, @Nullable Conjunction<? extends P
...
+        }
     }
 
+
```
krishnangovindraj, line 174: wild space

krishnangovindraj (approved):
Looks good to me

flyingsilverfin (approved):

PR #307, opened 2023-11-09
Technical debt: improve error_messages, cleanup

dmitrii-ubskii:

`error_messages!` now accepts struct enum variants, rather than tuple variants. This forces the user to name the fields and to refer to the fields by name in the format strings, reducing user error.

- `error_messages!` overhaul
- minor code style cleanup: naming, functional style fixes, simplified structure
    - split `Variable::Anonymous(Visible)` and `_::Anonymous(Invisible)` to `Variable::Anonymous` and `_::Hidden` respectively.

### DIFF
--- a/rust/builder/mod.rs
+++ b/rust/builder/mod.rs
@@ -83,7 +83,7 @@ macro_rules! max {
         let args = [$($arg, )*];
         $crate::pattern::Expression::Function($crate::pattern::Function {
             function_name: $crate::common::token::Function::Max,
-            args: args.into_iter().map(|arg| Box::new(arg.into())).collect(),
+            args: args.into_iter().map(Into::into).collect(),
         })
     }}
 }
@@ -94,7 +94,7 @@ macro_rules! min {
         let args = [$($arg, )*];
         $crate::pattern::Expression::Function($crate::pattern::Function {
             function_name: token::Function::Min,
-            args: args.into_iter().map(|arg| Box::new(arg.into())).collect(),
+            args: args.into_iter().map(Into::into).collect(),
         })
     }}
 }
@@ -178,17 +178,17 @@ pub fn like<T: Into<String>>(value: T) -> Predicate {
 }
 
 pub fn abs<T: Into<Expression>>(arg: T) -> Function {
-    Function { function_name: token::Function::Abs, args: vec![Box::from(arg.into())] }
+    Function { function_name: token::Function::Abs, args: vec![arg.into()] }
 }
 
 pub fn ceil<T: Into<Expression>>(arg: T) -> Function {
-    Function { function_name: token::Function::Ceil, args: vec![Box::from(arg.into())] }
+    Function { function_name: token::Function::Ceil, args: vec![arg.into()] }
 }
 
 pub fn floor<T: Into<Expression>>(arg: T) -> Function {
-    Function { function_name: token::Function::Floor, args: vec![Box::from(arg.into())] }
+    Function { function_name: token::Function::Floor, args: vec![arg.into()] }
 }
 
 pub fn round<T: Into<Expression>>(arg: T) -> Function {
-    Function { function_name: token::Function::Round, args: vec![Box::from(arg.into())] }
+    Function { function_name: token::Function::Round, args: vec![arg.into()] }
 }
--- a/rust/builder/mod.rs
+++ b/rust/builder/mod.rs
@@ -33,37 +33,37 @@ use crate::{
 
 #[macro_export]
 macro_rules! typeql_match {
-    ($($pattern:expr),* $(,)?) => {{
-        $crate::query::MatchClause::from_patterns(vec![$($pattern.into()),*])
-    }}
+    ($($pattern:expr),* $(,)?) => {
+        $crate::query::MatchClause::new($crate::pattern::Conjunction::new(vec![$($pattern.into()),*]))
+    }
 }
 
 #[macro_export]
 macro_rules! typeql_insert {
-    ($($thing_statement:expr),* $(,)?) => {{
+    ($($thing_statement:expr),* $(,)?) => {
         $crate::query::TypeQLInsert::new(vec![$($thing_statement),*])
-    }}
+    }
 }
 
 #[macro_export]
 macro_rules! typeql_define {
-    ($($pattern:expr),* $(,)?) => {{
+    ($($pattern:expr),* $(,)?) => {
         $crate::query::TypeQLDefine::new(vec![$($pattern.into()),*])
-    }}
+    }
 }
 
 #[macro_export]
 macro_rules! typeql_undefine {
-    ($($pattern:expr),* $(,)?) => {{
+    ($($pattern:expr),* $(,)?) => {
         $crate::query::TypeQLUndefine::new(vec![$($pattern.into()),*])
-    }}
+    }
 }
 
 #[macro_export]
 macro_rules! and {
-    ($($pattern:expr),* $(,)?) => {{
+    ($($pattern:expr),* $(,)?) => {
         $crate::pattern::Conjunction::new(vec![$($pattern.into()),*])
-    }}
+    }
 }
 
 #[macro_export]
@@ -80,10 +80,9 @@ macro_rules! or {
 #[macro_export]
 macro_rules! max {
     ($($arg:expr),* $(,)?) => {{
-        let args = [$($arg, )*];
         $crate::pattern::Expression::Function($crate::pattern::Function {
             function_name: $crate::common::token::Function::Max,
-            args: args.into_iter().map(Into::into).collect(),
+            args: vec![$($arg.into()),*],
         })
     }}
 }
@@ -91,10 +90,9 @@ macro_rules! max {
 #[macro_export]
 macro_rules! min {
     ($($arg:expr),* $(,)?) => {{
-        let args = [$($arg, )*];
         $crate::pattern::Expression::Function($crate::pattern::Function {
             function_name: token::Function::Min,
-            args: args.into_iter().map(Into::into).collect(),
+            args: vec![$($arg.into()),*],
         })
     }}
 }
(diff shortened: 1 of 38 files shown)


rust/builder/mod.rs
```diff
@@ -33,83 +33,82 @@ use crate::{
 
 #[macro_export]
 macro_rules! typeql_match {
-    ($($pattern:expr),* $(,)?) => {{
-        $crate::query::MatchClause::from_patterns(vec![$($pattern.into()),*])
-    }}
+    ($($pattern:expr),* $(,)?) => {
+        $crate::query::MatchClause::new($crate::pattern::Conjunction::new(vec![$($pattern.into()),*]))
+    }
```
dmitrii-ubskii, line 38: Simplifies `MatchClause` methods

rust/builder/mod.rs
```diff
@@ -33,83 +33,81 @@ use crate::{
...
-        }
-    }}
+    ($pattern:expr $(,)?) => { 
+        compile_error!("Useless disjunction of one pattern");
+    };
+
+    ($($pattern:expr),+ $(,)?) => {
+        $crate::pattern::Disjunction::new(vec![$($pattern.into()),*])
+    };
 }
```
dmitrii-ubskii, line 78: This change allows us to actually return a `Disjunction` rather than a generic `Pattern`, and avoids confusion where `or!(x)` would simply return `x` rather than a disjunction containing `x`.

rust/common/error/macros.rs
```diff
@@ -24,19 +24,19 @@
 macro_rules! error_messages {
     {
         $name:ident code: $code_pfx:literal, type: $message_pfx:literal,
-        $($error_name:ident( $($inner:ty),* $(,)? ) = $code:literal: $body:literal),+ $(,)?
+        $($error_name:ident $({ $($field:ident : $inner:ty),+ $(,)? })? = $code:literal: $body:literal),+ $(,)?
     } => {
```
dmitrii-ubskii, line 28: Error messages are now struct variants rather than tuple variants. This change allows us to name the payload, refer to it by name in the formatting string, and greatly simplifies this macro.

rust/parser/mod.rs
```diff
@@ -64,18 +64,6 @@ pub(crate) struct TypeQLParser;
...
-    ($($arg:tt)*) => {
-        debug_assert_eq!($($arg)*, "\n{}", line!())
-    };
-}
-
-macro_rules! dbg_assert_line {
-    ($($arg:tt)*) => {
-        debug_assert!($($arg)*, "\n{}", line!())
-    };
-}
```
dmitrii-ubskii, line 77: `debug_assert!` already prints the line it's on.

rust/pattern/constraint/type_/owns.rs
```diff
@@ -108,21 +108,12 @@ impl From<TypeReference> for OwnsConstraint {
...
-    fn from((attribute_type, overridden_attribute_type): (Label, Label)) -> Self {
-        OwnsConstraint::from((TypeReference::Label(attribute_type), TypeReference::Label(overridden_attribute_type)))
+impl<T: Into<Label>, U: Into<Label>> From<(T, U)> for OwnsConstraint {
+    fn from((attribute_type, overridden_attribute_type): (T, U)) -> Self {
+        OwnsConstraint::from((
+            TypeReference::Label(attribute_type.into()),
+            TypeReference::Label(overridden_attribute_type.into()),
+        ))
     }
 }
```
dmitrii-ubskii, line 118: Simplified some of the conversions

rust/pattern/expression/function.rs
```diff
@@ -28,7 +28,7 @@ use crate::{common::token, pattern::LeftOperand, variable::variable::VariableRef
 #[derive(Debug, Clone, Eq, PartialEq)]
 pub struct Function {
     pub(crate) function_name: token::Function,
-    pub(crate) args: Vec<Box<Expression>>,
+    pub(crate) args: Vec<Expression>,
```
dmitrii-ubskii, line 31: `Vec<>` is already on the heap, `Box<>` here is only harmful.

rust/pattern/test/mod.rs
```diff
@@ -49,7 +49,6 @@ get;"#;
             and!(cvar("com").has(("name", cvar("n1"))), cvar("n1").eq("the-company"), cvar("com").isa("company"),),
             and!(cvar("com").has(("name", cvar("n2"))), cvar("n2").eq("another-company"), cvar("com").isa("company"),)
         )
-        .into_disjunction()
```
dmitrii-ubskii, line 52: `or!()` now returns a `Disjunction` directly.

rust/query/match_clause.rs
```diff
@@ -46,10 +46,6 @@ impl MatchClause {
         Self { conjunction }
     }
 
-    pub fn from_patterns(patterns: Vec<Pattern>) -> Self {
-        Self::new(Conjunction::new(patterns))
-    }
```
dmitrii-ubskii, line 51: This was only used in the builder and is easily inlined by simply repeating the body of this function.

rust/query/typeql_fetch.rs
```diff
@@ -147,24 +149,22 @@ impl<T: Into<Variable>> From<T> for ProjectionKeyVar {
 }
 
 #[derive(Debug, Eq, PartialEq)]
-pub enum ProjectionKeyLabel {
-    Quoted(String),
-    Unquoted(String),
+pub struct ProjectionKeyLabel {
+    pub label: String,
 }
```
dmitrii-ubskii, line 154: `ProjectionKeyLabel` is only quoted on `to_string()` and only when it has to be.

rust/variable/variable_concept.rs
```diff
@@ -28,73 +28,56 @@ use crate::{
...
-}
-
 #[derive(Debug, Clone, Eq, Hash, PartialEq)]
 pub enum ConceptVariable {
-    Anonymous(Visibility),
-    Name(String),
+    Anonymous,
+    Hidden,
+    Named(String),
 }
```
dmitrii-ubskii, line 36: Simplified

rust/variable/variable_concept.rs
```diff
@@ -28,73 +28,56 @@ use crate::{
...
-        ConceptVariable::Name(name)
-    }
-
-    pub fn anonymous() -> ConceptVariable {
-        ConceptVariable::Anonymous(Visibility::Visible)
-    }
-
-    pub fn hidden() -> ConceptVariable {
-        ConceptVariable::Anonymous(Visibility::Invisible)
-    }
```
dmitrii-ubskii, line 56: These are not necessary, the variants can be constructed directly with near-identical syntax.

rust/variable/variable.rs
```diff
@@ -41,10 +41,10 @@ impl Variable {
         }
     }
 
-    pub fn is_name(&self) -> bool {
+    pub fn is_named(&self) -> bool {
```
dmitrii-ubskii, line 44: `is_name()` is from the now removed `Reference`. `is_named()` makes more sense on a `Variable`.

rust/query/typeql_fetch.rs
flyingsilverfin: This is actually not quite the intent - the user can choose to quote things even if they don't have to be and we're trying to remember which way they wrote the syntax. The `parser` always used quoted if the user wrote quotes. 

The 'choosing' of whether we have to quote or not only happens in the programmatic API where the user doesn't have to quote their strings (obviously!)

rust/variable/variable.rs
flyingsilverfin: Ok, this is fine

rust/variable/variable_concept.rs
flyingsilverfin: We can make this change, but we have to make clear we're changing a fundamental model of TypeQL: We used to have 'anonymous' and 'named' variables only - now we're introducing a third fundamental type of variable: 'hidden'.

I tried to think of further implications - only came up with that we will have to decide how we want to treat them on the backend later. We can also split them on the backend and use numbers `$1` for example for hidden, and `$_` for anonymous in the future. Otherwise we may want to retract this change but we can come back here

flyingsilverfin: A big conceptual change like this also has to be outlined clearly in the PR description.

dmitrii-ubskii: This is not a big conceptual or structural change. The change is 1-to-1 from `Anonymous(Visible)` and `Anonymous(Invisible)` to `Anonymous` and `Hidden` respectively.

rust/query/typeql_fetch.rs
dmitrii-ubskii: We need to pick a lane here. Either we parse to have a faithful representation of the user input, or we parse to have a consistent internally usable representation of the user's query. We are currently stuck in the middle, and I would argue that we should stick to the latter, as the main two purposes for the typeql library are to parse the queries on the server side, and to build the queries on the client side, neither of which should care about faithfully representing the raw input.

Here, not having to worry about whether the user happened to quote the label saves us a match on the enum anywhere we might care to get that label. We can simplify a lot of things like these, and the benefits will add up.

rust/variable/variable_concept.rs
flyingsilverfin: it is a big conceptual change - we had 2 types of variables, now we have 3, rather than 2 where one has 2 variations (invisible/visible). This kind of thing affects how we think of query structures in our heads, so it is important.

Example - a conjunction in my head is always made of variables. Any hidden variable is anonymous, because anonymous variables can't be referred to twice. Thus any generated variable is anonymous, and our conjunctions are made of named or anonymous variables only. We can abstract away having another type of variable that is generated.

We could discuss that we should be treating generated variables as a different _kind_ of variable, has you've done here. This is in my mind OK.

flyingsilverfin (approved):

PR #313, opened 2024-01-25
Merge typedb-common repository into typeql

dmitrii-ubskii:

As part of the effort to reduce the number of vaticle organization repositories, we merge typedb-common into the typeql repo as a subpackage.


common/java/BUILD
```diff
@@ -0,0 +1,106 @@
...
+
+deploy_github(
+    name = "deploy-github",
+    organisation = deployment_github['github.organisation'],
+    repository = deployment_github['github.repository'],
+    title = "TypeDB Common",
+    title_append_version = True,
+    release_description = "//:RELEASE_NOTES_LATEST.md",
+    draft = False,
+)
```
flyingsilverfin, line 59: Also don't need this - is in the root

common/java/BUILD
```diff
@@ -0,0 +1,106 @@
...
+        "concurrent/*",
+        "concurrent/actor/*.java",
+        "concurrent/actor/eventloop/*.java",
+        "deployment.bzl",
+        "exception/*",
+        "util/*",
+        "yaml/*.java"
+    ]),
+    license_type = "apache-header",
+)
```
flyingsilverfin, line 93: Check these files are still correct (eg. deployment shouldn't be here)

flyingsilverfin (changes requested):
Also please double make sure you notify the team so they know that these files have a new home

flyingsilverfin (approved):

PR #317, opened 2024-02-14
Allow variables to have a leading digit

flyingsilverfin:
## Usage and product changes

We modify the behaviour of #310 which unified variables and labels to have the same valid identifier syntax, which eliminated the capability of variables to have a leading number. For example, the variable `$0` was banned.

This PR reverts this specific behaviour, and enables usage of variables with leading digits:
```
match
$1_a isa entity;
get;
```
is made valid again.

Testing specified in https://github.com/vaticle/typedb-behaviour/pull/281

## Implementation

We split the identifier regexes and grammar rules that validated labels and variables before, to be specific to `Variable` and `Label` identifiers.


grammar/TypeQL.g4
```diff
@@ -332,35 +332,46 @@ DATETIME_       : DATE_FRAGMENT_ 'T' TIME_              ;
 
 VAR_CONCEPT_            : VAR_CONCEPT_ANONYMOUS_ | VAR_CONCEPT_NAMED_   ;
 VAR_CONCEPT_ANONYMOUS_  : '$_' ;
-VAR_CONCEPT_NAMED_      : '$'  TYPE_CHAR_H_ TYPE_CHAR_T_* ;
-VAR_VALUE_              : '?'  TYPE_CHAR_H_ TYPE_CHAR_T_* ;
+VAR_CONCEPT_NAMED_      : '$'  IDENTIFIER_VAR_H IDENTIFIER_VAR_T* ;
+VAR_VALUE_              : '?'  IDENTIFIER_VAR_H IDENTIFIER_VAR_T* ;
 IID_                    : '0x' [0-9a-f]+                  ;
-LABEL_                  : TYPE_CHAR_H_ TYPE_CHAR_T_*      ;
+LABEL_                  : IDENTIFIER_LABEL_H IDENTIFIER_LABEL_T*      ;
```
flyingsilverfin, line 338: `TYPE_` prefix was a hangover from the past when "label" was called "type" in the grammar. Fixed here.

rust/common/identifier.rs
```diff
@@ -23,10 +23,7 @@ use std::sync::OnceLock;
 
 use regex::{Regex, RegexBuilder};
 
-pub fn is_valid_identifier(identifier: &str) -> bool {
-    static REGEX: OnceLock<Regex> = OnceLock::new();
-    let regex = REGEX.get_or_init(|| {
-        let identifier_start = "A-Za-z\
+const IDENTIFIER_CHAR: &str = "A-Za-z\
```
flyingsilverfin, line 26: Refactored constant strings out here, and only compute the regexes once in the `OnceLock`.

grammar/TypeQL.g4
```diff
@@ -332,35 +332,46 @@ DATETIME_       : DATE_FRAGMENT_ 'T' TIME_              ;
...
+                               | IDENTIFIER_CHAR_
+                               ;
+fragment IDENTIFIER_VAR_T      : IDENTIFIER_VAR_H
+                               | IDENTIFIER_CONNECTOR_
+                               ;
+fragment IDENTIFIER_LABEL_H    : IDENTIFIER_CHAR_ ;
+fragment IDENTIFIER_LABEL_T    : IDENTIFIER_LABEL_H
+                               | IDENTIFIER_DIGIT_
+                               | IDENTIFIER_CONNECTOR_
+                               ;
```
dmitrii-ubskii, line 373: I think this would be easier to read if `_T`s explicitly listed the character groups. This would also align with the regexes in the code.

rust/common/identifier.rs
```diff
@@ -38,16 +35,28 @@ pub fn is_valid_identifier(identifier: &str) -> bool {
...
+        RegexBuilder::new(&identifier_pattern).build().unwrap()
+    });
+    regex.is_match(identifier)
+}
+
+pub fn is_valid_var_identifier(identifier: &str) -> bool {
+    static REGEX: OnceLock<Regex> = OnceLock::new();
+    let regex = REGEX.get_or_init(|| {
+        let identifier_start = format!("{}{}", IDENTIFIER_CHAR, IDENTIFIER_DIGIT);
+        let identifier_tail = format!("{}{}{}", IDENTIFIER_CHAR, IDENTIFIER_DIGIT, IDENTIFIER_CONNECTOR);
```
dmitrii-ubskii, line 59: `_head` and `_tail`

grammar/TypeQL.g4
dmitrii-ubskii: These should also have a trailing underscore like all other fragments. `.pest` follows this convention already.

rust/parser/typeql.pest
```diff
@@ -354,37 +354,42 @@ DATETIME_ = @{ DATE_FRAGMENT_ ~ "T" ~ TIME_ ~ WB }
 VAR_ = ${ VAR_CONCEPT_ | VAR_VALUE_ }
 VAR_CONCEPT_ = @{ VAR_CONCEPT_ANONYMOUS_ | VAR_CONCEPT_NAMED_ }
 VAR_CONCEPT_ANONYMOUS_ = @{ "$_" ~ WB }
-VAR_CONCEPT_NAMED_ = @{ "$" ~ IDENTIFIER }
-VAR_VALUE_ = @{ "?" ~ IDENTIFIER }
+VAR_CONCEPT_NAMED_ = @{ "$" ~ IDENTIFIER_VAR_H_ ~ IDENTIFIER_VAR_T_*  ~ WB }
+VAR_VALUE_ = @{ "?" ~ (IDENTIFIER_VAR_H_ ~ IDENTIFIER_VAR_T_* ~ WB) }
```
dmitrii-ubskii, line 358: `IDENTIFIER_VAR_` exists already

rust/parser/typeql.pest
```diff
@@ -354,37 +354,42 @@ DATETIME_ = @{ DATE_FRAGMENT_ ~ "T" ~ TIME_ ~ WB }
...
 IID_ = @{ "0x" ~ ASCII_HEX_DIGIT+ ~ WB }
-LABEL_ = @{ IDENTIFIER }
-LABEL_SCOPED_ = @{ IDENTIFIER ~ ":" ~ IDENTIFIER }
-IDENTIFIER = @{ TYPE_CHAR_H_ ~ TYPE_CHAR_T_* ~ WB }
+LABEL_ = @{ (IDENTIFIER_LABEL_H_ ~ IDENTIFIER_LABEL_T_* ~ WB) }
+LABEL_SCOPED_ = @{ (IDENTIFIER_LABEL_H_ ~ IDENTIFIER_LABEL_T_*) ~ ":" ~ (IDENTIFIER_LABEL_H_ ~ IDENTIFIER_LABEL_T_*)  ~ WB }
```
dmitrii-ubskii, line 361: We should probably have an `IDENTIFIER_LABEL_` rule.

dmitrii-ubskii: Although the ANTLR grammar reuses `LABEL_` instead. If we do that, `IDENTIFIER_VAR_` might have to go as well. I'd rather we introduce both in ANTLR though.

dependencies/vaticle/repositories.bzl
```diff
@@ -27,6 +27,6 @@ def vaticle_dependencies():
 def vaticle_typedb_behaviour():
     git_repository(
         name = "vaticle_typedb_behaviour",
-        remote = "https://github.com/vaticle/typedb-behaviour",
-        commit = "031ab431f14f86b0ceed3b12bed8267f88868c72", # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_typedb_behaviour
+        remote = "https://github.com/flyingsilverfin/typedb-behaviour",
+        commit = "d8790a952d04948edf9ca9490070b133862c445e", # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_typedb_behaviour
     )
```
dmitrii-ubskii, line 32: Note

dmitrii-ubskii (changes requested):

rust/parser/typeql.pest
flyingsilverfin: I just canceled them all, so we don't end up with confusion re. `~wb`

dmitrii-ubskii (approved):

PR #345, opened 2024-07-24
TypeQL 3 grammar enhancements

dmitrii-ubskii:
## Usage and product changes

1. New rich `fetch` syntax (see below).
2. Standardise the vocabulary of `pipeline`, `stage`, `clause`, `operator`. A _pipeline_ consists of _stages_. Each _stage_ may be an _operator_, which modifies the data stream without accessing the database (e.g. `count`, `mean($x)`), or a _clause,_ which may fetch data from the database to modify the stream (e.g. `match`, `fetch`).
3. `list()` stream reduce operator.
4. `$x in [$a, $b, $c]` and other list expressions now allowed in `in`-statements (previously stream-only)

### New `fetch` syntax sample

```php
... # incoming pipeline
fetch {
# Printing values directly from pipeline
  "key_1": $x, # var $x (from input stream) holds a value or list
  "key_2": <EXPR>, # <EXPR> is an expression like $x + $y

# Inline attribute retrieval variations
  "key_3": $y.attr, # var $y holds an object with singleton attribute 'attr'
  "key_4": [ $y.attr ], # object var $y has multiple attributes 'attr'
  "key_5": $y.attr[], # object var $y has a list attribute 'attr'

# Function call variations
  "key_6": my_fun1($x,$y), # function my_fun1 has single-return
  "key_7": [ my_fun2($x,$y) ], # function my_fun2 has stream-return

# Match-fetch subqueries
  "key_8": [
    match ...;
    fetch {
      "sub_key": $z, 
      ...
    };
  ]

# Match-reduce-value subqueries
  "key_9": 
    match ...;
    reduce agg($z); # agg could be, e.g., 'count', 'sum', or 'list'

# Nested keys: Nothing stops you from nesting the above!
  "super_key": {
    "sub_key_1": $x,
    "sub_key_2": $y.attr,
    "sub_key_3": [
      ... # some subquery
    ]
  }
};


```


rust/parser/typeql.pest
```diff
@@ -196,14 +198,14 @@ definable = { ( definition_type ~ SEMICOLON ) | definition_function | definition
 
 // TYPE DEFINITION =============================================================
 
-definition_type = { kind? ~ identifier ~ type_capability ~ ( COMMA ~ type_capability )* }
+definition_type = { kind? ~ label ~ ( type_capability ~ ( COMMA ~ type_capability )* )? }
```
dmitrii-ubskii, line 201: `define entity person;` replaces `define person sub entity;`, meaning capabilities are now optional.

rust/common/error/mod.rs
```diff
@@ -101,7 +101,7 @@ error_messages! { TypeQLError
     InvalidAttributeTypeRegex { regex: String } =
         12: "Invalid regular expression '{regex}'.",
     GetVarRepeating { variable: Variable } =
-        13: "The variable '{variable}' occurred more than once in get query filter.",
+        13: "The variable '{variable}' occurred more than once in get query select.",
```
flyingsilverfin, line 104: `get` is obsolete - can you do a quick search for occurences please?

rust/common/error/mod.rs
```diff
@@ -117,7 +117,7 @@ error_messages! { TypeQLError
     InsertModifiersRequireMatch { insert: String } =
         20: "The insert query '{insert}' contains query modifiers that require a 'match' clause be specified",
     VariableNotNamed =
-        21: "Anonymous variable encountered in a match query filter.",
+        21: "Anonymous variable encountered in a match query select.",
```
flyingsilverfin, line 120: Also we should use the new terminology, such as `select operator` - there is no 'match query' anymore :)

rust/parser/typeql.pest
```diff
@@ -551,30 +560,31 @@ CURLY_CLOSE = _{ "}" }
...
-                   | '\u{00F8}'..'\u{02FF}'
-                   | '\u{0370}'..'\u{037D}'
-                   | '\u{037F}'..'\u{1FFF}'
-                   | '\u{200C}'..'\u{200D}'
-                   | '\u{2070}'..'\u{218F}'
-                   | '\u{2C00}'..'\u{2FEF}'
-                   | '\u{3001}'..'\u{D7FF}'
-                   | '\u{F900}'..'\u{FDCF}'
-                   | '\u{FDF0}'..'\u{FFFD}'
+                   // | '\u{00C0}'..'\u{00D6}'
```
flyingsilverfin, line 571: Why are these commented?

flyingsilverfin (changes requested):

rust/common/error/mod.rs
dmitrii-ubskii: Fixed. No other matches found.

rust/parser/typeql.pest
dmitrii-ubskii: That was not meant to be pushed, fixed.
That was for the autogenerated tests, so that the printed generated queries which failed would be somewhat readable.

flyingsilverfin (approved):

rust/parser/typeql.pest
```diff
@@ -132,8 +133,11 @@ statement_single = { statement_is | statement_comparison | statement_assignment
 statement_is = { var ~ IS ~ var }
 statement_comparison = { expression_value ~ comparison }
 statement_assignment = { assignment_left ~ ASSIGN ~ expression }
-assignment_left = { vars | struct_destructor }
-statement_in = { vars ~ IN ~ expression_function }
+assignment_left = { vars_assignment | struct_destructor }
+statement_in = { vars_assignment ~ IN ~ ( expression_function | expression_list ) }
+
+vars_assignment = { var_assignment ~ ( COMMA ~ var_assignment )* }
+var_assignment = { var_optional | var }
```
krishnangovindraj, line 140: anchor

PR #364, opened 2024-09-12
Rename override to specialise. Remove specialisations for owns and plays

farost:
## Usage and product changes
Rename override to specialise. Remove specialisations for owns and plays respecting the server's changes: https://github.com/typedb/typedb/pull/7157

### DIFF
--- a/rust/parser/typeql.pest
+++ b/rust/parser/typeql.pest
@@ -90,12 +90,12 @@ sub_constraint = { SUB_ ~ type_ref_any }
 value_type_constraint = { VALUE ~ value_type }
 label_constraint = { LABEL ~ ( label_scoped | label ) }
 owns_constraint = { OWNS ~ type_ref_list
-                  | OWNS ~ type_ref ~ ( AS ~ type_ref )?
+                  | OWNS ~ type_ref
                   }
 relates_constraint = { RELATES ~ type_ref_list
                      | RELATES ~ type_ref ~ ( AS ~ type_ref )?
                      }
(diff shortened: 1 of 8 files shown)


rust/parser/typeql.pest
```diff
@@ -263,12 +263,12 @@ undefinable = { undefine_from | undefine_function | undefine_struct | label }
...
                 | undefine_annotation_from_type
                 | undefine_capability
-                | undefine_override
+                | undefine_specialise
                 }
 undefine_annotation_from_capability = { annotation_category ~ FROM ~ label ~ type_capability_base }
 undefine_annotation_from_type = { annotation_category ~ FROM ~ label }
 undefine_capability = { type_capability_base ~ FROM ~ label }
-undefine_override = { AS ~ label ~ FROM ~ label ~ type_capability_base }
+undefine_specialise = { AS ~ label ~ FROM ~ label ~ type_capability_base }
```
farost, line 271: Ideally, there should be only `relates`, but, as I understand, we prefer to use `capability` in general in these situations (parsing for pure `relates` declarations is private, and it previously combined all the capabilities instead of just 3: `owns`, `relates`, `plays`). So I'm going to leave it here and just add errors on the server side ("You tried to undefine specialisation for `owns`/`sub`/..., but this feature doesn't exist for this capability).

flyingsilverfin: Use `pub(in crate::parser)`

farost: Done

PR #365, opened 2024-09-13
Expose some extra fields in statements

krishnangovindraj:
## Usage and product changes
Expose some more required fields.

## Implementation
Expose some more required fields.


rust/statement/thing/mod.rs
```diff
@@ -135,24 +135,22 @@ impl fmt::Display for RolePlayer {
 #[derive(Debug, Clone, Eq, PartialEq)]
 pub struct AttributeValueStatement {
     span: Option<Span>,
-    pub type_: Option<TypeRef>,
+    pub var: Variable,
```
krishnangovindraj, line 138: @dmitrii-ubskii , I'm probably missing something but why was this `Option<TypeRef>` ?

PR #367, opened 2024-09-23
Implement duration literal parsing

dmitrii-ubskii:
# Release notes: product changes

We now accept the ISO 8601 duration literal format (`P3Y6M4DT12H30M5.346S`) in TypeQL queries.

### DIFF
--- a/rust/typeql.rs
+++ b/rust/typeql.rs
@@ -9,7 +9,6 @@
 #![deny(rust_2024_compatibility)]
 #![deny(elided_lifetimes_in_paths)]
 #![deny(unused_must_use)]
-
 #![allow(edition_2024_expr_fragment_specifier)]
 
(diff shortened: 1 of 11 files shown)


rust/typeql.rs
```diff
@@ -10,6 +10,8 @@
 #![deny(elided_lifetimes_in_paths)]
 #![deny(unused_must_use)]
 
+#![allow(edition_2024_expr_fragment_specifier)]
```
dmitrii-ubskii, line 13: `:expr` in macros will accept `const` blocks and `_` expressions in the 2024 edition. We don't currently depend on any specific behaviour in our macros.

(https://doc.rust-lang.org/nightly/edition-guide/rust-2024/macro-fragment-specifiers.html#migration)

PR #374, opened 2024-10-30
Make Is statement fields public

dmitrii-ubskii:
## Usage and product changes

We make the `lhs` and `rhs` fields of the `Is` statement struct public for use in typedb core.


krishnangovindraj (approved):
Cool

PR #375, opened 2024-10-31
Add unescape version for regex annotations

farost:
## Usage and product changes
We add a separate `unescape` method for strings inside `regex` annotations to preserve regex escaped characters and unescape `"` quotes.

### DIFF
--- a/rust/value.rs
+++ b/rust/value.rs
@@ -324,23 +324,19 @@ impl fmt::Display for StructLiteral {
 
 impl StringLiteral {
     pub fn unescape(&self) -> Result<String> {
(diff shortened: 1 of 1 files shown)


rust/value.rs
```diff
@@ -343,25 +372,10 @@ impl StringLiteral {
                         full_string: escaped_string.to_owned(),
                         escape: String::from(r"\"),
                     }
-                    .into());
+                        .into());
```
dmitrii-ubskii, line 375: `fmt`

dmitrii-ubskii (approved):

flyingsilverfin (approved):

PR #379, opened 2024-11-22
Add function redefine, other small changes for functions

krishnangovindraj:
## Usage and product changes
Adds function redefine syntax, and some small changes.

## Implementation
* Add function redefine
* Makes some fields public
* Functions keep their unparsed string representation so the function manager can store it.

### DIFF
--- a/rust/schema/definable/function.rs
+++ b/rust/schema/definable/function.rs
@@ -258,13 +258,13 @@ impl fmt::Display for ReturnStatement {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
             ReturnStatement::Stream(return_stream) => {
-                write!(f, "{} {}", token::Keyword::Return, return_stream)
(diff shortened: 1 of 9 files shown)


rust/schema/definable/function.rs
```diff
@@ -20,11 +23,12 @@ pub struct Function {
     span: Option<Span>,
     pub signature: Signature,
     pub block: FunctionBlock,
+    pub unparsed: String,
```
krishnangovindraj, line 26: Flag a hack.
I'm supposed to use the span for this. But then I have to pass the unparsed query in the core functions.
I'm not sure whether I want to do that today or leave this in as a hack.

krishnangovindraj (commented):
Flag the hack

rust/schema/definable/function.rs
krishnangovindraj: We're happy with this for now.

PR #382, opened 2024-11-27
Low hanging optimisations for TypeQL

krishnangovindraj:
## Usage and product changes
Low hanging optimisations for TypeQL

## Implementation
* The grammar no longer validates that identifiers are not reserved. The `Identifier::as_str` function with `as_str_unreserved` and `as_str_unchecked` .
* Spans hold the line & column information, since storing this eagerly involved traversing the string. Instead, we store the offset and can re-derive line & column given the string.

### DIFF
--- a/rust/common/identifier.rs
+++ b/rust/common/identifier.rs
@@ -26,7 +26,7 @@ impl Identifier {
         if !is_reserved_keyword(&self.ident) {
             Ok(&self.ident)
         } else {
-            Err(TypeQLError::ReservedKeywordAsIdentifier { identifier: self.ident.to_owned() })
+            Err(TypeQLError::ReservedKeywordAsIdentifier { identifier: self.clone() })
         }
     }
 
--- a/rust/common/identifier.rs
+++ b/rust/common/identifier.rs
@@ -8,10 +8,7 @@ use std::{fmt, sync::OnceLock};
 
 use regex::{Regex, RegexBuilder};
 
-use crate::{
-    common::{Span, Spanned},
-    pretty::Pretty,
-};
+use crate::{common::{Span, Spanned}, is_reserved_keyword, pretty::Pretty};
 
 #[derive(Debug, Clone, Eq, PartialEq, Hash)]
 pub struct Identifier {
@@ -25,6 +22,7 @@ impl Identifier {
     }
 
     pub fn as_str(&self) -> &str {
+        debug_assert!(!is_reserved_keyword(&self.ident));
         &self.ident
     }
 }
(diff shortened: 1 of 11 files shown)


rust/common/mod.rs
```diff
@@ -22,40 +22,39 @@ pub struct LineColumn {
...
+    pub end_offset: usize,
 }
 
 pub trait Spanned {
     fn span(&self) -> Option<Span>;
 }
 
-pub trait DisplaySpanned: Spanned {
-    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result;
+pub trait Spannable {
```
krishnangovindraj, line 35: Was `ExtractSpan`. Then I introduced `line_col`and renamed it.

rust/common/identifier.rs
```diff
@@ -24,7 +25,15 @@ impl Identifier {
...
-    pub fn as_str(&self) -> &str {
+    pub fn as_str_unreserved(&self) -> Result<&str, TypeQLError> {
+        if !is_reserved_keyword(&self.ident) {
+            Ok(&self.ident)
+        } else {
+            Err(TypeQLError::ReservedKeywordAsIdentifier { identifier: self.clone() })
+        }
+    }
+
+    pub fn as_str_unchecked(&self) -> &str {
```
krishnangovindraj, line 36: New functions

krishnangovindraj (commented):
```
string_enum! { Keyword
    Abstract = "abstract",
    As = "as",
    Alias = "alias",
    Assign = "=",
    Check = "check",
    First = "first",
    From = "from",
    Fun = "fun",
    Has = "has",
    IID = "iid",
    In = "in",
    Is = "is",
    Isa = "isa",
    IsaX = "isa!",
    Label = "label",
    Last = "last",
    Links = "links",
    Not = "not",
    Of = "of",
    Or = "or",
    Owns = "owns",
    Plays = "plays",
    Relates = "relates",
    Return = "return",
    Struct = "struct",
    Sub = "sub",
    SubX = "sub!",
    Try = "try",
    Value = "value",
    Within = "within",
}
```
This might be too tight a list of keywords to ban.

rust/common/mod.rs
```diff
@@ -22,40 +22,39 @@ pub struct LineColumn {
...
+    fn extract(&self, span: Span) -> &str;
+
+    fn line_col(&self, span: Span) -> Option<(LineColumn, LineColumn)>;
```
flyingsilverfin, line 38: Nice, we can keep the API now

flyingsilverfin (approved):

PR #389, opened 2025-01-22
Add 'dec' suffix to decimal literals

krishnangovindraj:
## Usage and product changes
Require decimal literals to be suffixed with 'dec'

## Implementation
Updates grammar & datastructures.

### DIFF
--- a/rust/parser/test/match_queries.rs
+++ b/rust/parser/test/match_queries.rs
@@ -492,9 +492,7 @@ $type relates someRole;"#;
 fn test_parsing_double_decimal_literal() {
     let query = r#"match
 let $x = -5.0 + -4.0dec;"#;
-
     let parsed = parse_query(query).unwrap();
-    // let expected = match_!(var("x").assign(5), var("a").equals(var("x")).isa("age"));
-
+    // let expected = todo!("When typeql builders exist");
     assert_valid_eq_repr!(expected, parsed, query);
 }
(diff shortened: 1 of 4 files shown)


rust/value.rs
```diff
@@ -44,6 +44,12 @@ pub struct SignedIntegerLiteral {
     pub integral: String,
 }
 
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct SignedDoubleLiteral {
+    pub sign: Option<Sign>,
+    pub double: String,
```
krishnangovindraj, line 50: TODO: I should strip out the suffix in this string.

rust/parser/test/match_queries.rs
```diff
@@ -487,3 +487,14 @@ $type relates someRole;"#;
...
     assert_valid_eq_repr!(expected, parsed, query);
 }
+
+#[test]
+fn test_parsing_double_decimal_literal() {
+    let query = r#"match
+let $x = -5.0 + -4.0dec;"#;
+
+    let parsed = parse_query(query).unwrap();
+    // let expected = match_!(var("x").assign(5), var("a").equals(var("x")).isa("age"));
```
farost, line 497: How do these tests work now? I see other tests having `expected` vars commented out

rust/value.rs
farost: It sounds like a `debug_assert` to me, haha. But it would require you to write a `new` method
Whatever

farost (approved):
Looks good to me, I'll check your explanation about tests later

PR #391, opened 2025-01-28
Enhance error printing and query source span availability

flyingsilverfin:
## Usage and product changes

We improve the error messages to show a `^` column indicator along with `-->` line indicator:
```
        define
        attribute name value string;
-->     entity person owns name @range(0..10);
                                ^
```

We also expose more information about where in the original query spans which sourced various internal data structures.

## Implementation

* Improve error message printing
* Include further information about `Span`s throughout TypeQL data structures

### DIFF
--- a/rust/common/mod.rs
+++ b/rust/common/mod.rs
@@ -91,7 +91,10 @@ impl Spannable for &str {
             .collect();
         if annotated {
             let lines_start = line.checked_sub(lines_before).unwrap_or(0);
-            let lines_end = min(lines.len(), line.checked_add(lines_after.checked_add(1).unwrap_or(usize::MAX)).unwrap_or(usize::MAX));
+            let lines_end = min(
+                lines.len(),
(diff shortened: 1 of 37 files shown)


rust/common/mod.rs
```diff
@@ -51,6 +65,41 @@ impl Spannable for &str {
...
+        let lines: Vec<_> = self
+            .lines()
+            .enumerate()
+            .map(|(i, line_string)| {
+                if i == line {
+                    annotated = true;
+                    format!(
+                        "{SYNTAX_ANNOTATED_INDICATOR_LINE}{line_string}\n{}{SYNTAX_ANNOTATED_INDICATOR_COL}",
+                        " ".repeat(SYNTAX_ANNOTATED_INDENT + col)
+                    )
```
flyingsilverfin, line 86: We add both the line indicator, and a following like with an `^` column indicator

PR #394, opened 2025-02-20
Allow duration literals without date component

dmitrii-ubskii:
## Usage and product changes

We modify the grammar to accept a `duration_literal` without a date component (e.g. `PT1S`) as per the standard. 

## Implementation

The valid duration literals are:

- `"P" ~ duration_weeks` 
- `"P" ~ duration_date`
- `"P" ~ duration_date ~ "T" ~ duration_time`
- `"PT" ~ duration_time`

The last option was missing from the grammar.

### DIFF
--- a/rust/parser/typeql.pest
+++ b/rust/parser/typeql.pest
@@ -557,7 +557,7 @@ date_literal = ${ date_fragment ~ WB }
 datetime_literal = ${ date_fragment ~ "T" ~ time ~ WB }
 datetime_tz_literal = ${ date_fragment ~ "T" ~ time ~ ( " " ~ iana_timezone | iso8601_timezone_offset ) ~ WB }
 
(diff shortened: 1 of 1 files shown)


rust/value.rs
```diff
@@ -328,6 +329,7 @@ impl fmt::Display for DurationLiteral {
                     Some(time) => write!(f, "T{time}")?,
                 }
             }
+            DurationLiteral::Time(time) => write!(f, "T{time}")?,
```
krishnangovindraj, line 332: Should this not be PT?

krishnangovindraj (approved):
Approved with one question, which I'll bring up on Discord

rust/value.rs
dmitrii-ubskii: `"P"` is written unconditionally at the start of formatting.

PR #405, opened 2025-04-22
Add query end marker

flyingsilverfin:

TypeQL query pipelines cannot be distinguished since query stages can be arbitrarily chained and concatenated. To resolve this, we introduce the `end;` marker. For example:
```
match ...
insert ...
end;
match ...
insert...
```

Is now correctly interpretable as two separate match-insert queries! Compared to before:
```
match ...
insert ...
match ...
insert ...
```

Where it was not clear if this is one match-insert-match-insert pipeline, two match-insert pipelines, or 4 separate queries, or anything in between!

For simplicity, any query can be terminated with an 'end;` marker, though it is redundant in for schema queries and 'fetch' queries:
```
define ...;
end; # redundant!

match ...
insert ...
end;  # not redundant!

match ...
insert ...
fetch { ... };
end;  # redundant!
```

## Implementation

We introduce the 'end' marker that allows terminating any query, even if it is a redundant marker.

The parsed `Query` type now also contains a boolean indicating whether the query was explicitly ended with an 'end;' marker.


rust/parser/typeql.pest
```diff
@@ -5,17 +5,15 @@
  */
 
 eof_query = { SOI ~ query ~ EOI }
+eof_query_prefix = { SOI ~ query }
 eof_label = { SOI ~ label ~ EOI }
 eof_definition_function = { SOI ~ definition_function ~ EOI }
 eof_definition_struct = { SOI ~ definition_struct ~ EOI }
 
-// WARN: do not include in the autogenerated tests
-eof_query_prefix_partial = { SOI ~ query ~ any_partial ~ EOI }
```
flyingsilverfin, line 13: Interestingly, we don't actually need this any_partial, we can just drop the EOI.

rust/parser/mod.rs
```diff
@@ -127,13 +127,12 @@ fn parse_single(rule: Rule, string: &str) -> Result<Node<'_>> {
...
 pub(crate) fn visit_query_prefix(string: &str) -> Result<(Query, usize)> {
-    let parsed = parse_single(Rule::eof_query_prefix_partial, string);
+    let parsed = parse_single(Rule::eof_query_prefix, string);
     match parsed {
         Ok(node) => {
             let mut children = node.into_children();
             let query = children.consume_expected(Rule::query);
-            let remaining = children.consume_expected(Rule::any_partial);
-            let end_of_query_index = remaining.span().unwrap().begin_offset;
+            let end_of_query_index = query.span().unwrap().end_offset;
```
flyingsilverfin, line 135: The problem is @dmitrii-ubskii  that we revert back to not being able to include comments and whitespace after. But I think we can live with that...

rust/parser/test/pipeline.rs
flyingsilverfin, line 1: New file to test pipelines and multi-query

rust/query/mod.rs
```diff
@@ -18,26 +19,66 @@ pub mod schema;
 
 #[derive(Debug, Eq, PartialEq)]
 pub enum Query {
-    Schema(SchemaQuery),
-    Pipeline(Pipeline),
+    Schema(SchemaQuery, bool),
```
flyingsilverfin, line 22: It's very useful for systems (like console) to know if the query is explicitly terminated!

rust/query/mod.rs
```diff
@@ -18,26 +19,66 @@ pub mod schema;
...
+    Pipeline(Pipeline, bool),
 }
 
-enum_getter! { Query
```
flyingsilverfin, line 25: Just write this out basically, plus some new methods

rust/parser/typeql.pest
dmitrii-ubskii: Where do we use the partial parsing in the first place?

flyingsilverfin: In console - we either get a user writing a query interactively, in which we wait to get a double newline before submitting (here we wouldn't need the prefix parsing). However, we sometimes get copy-pasted scripts:
```
transaction write test
match ...;
insert ...;

commit
```
Where we don't want the commit to be parsed as part of the input

rust/parser/typeql.pest
```diff
@@ -5,17 +5,15 @@
  */
 
 eof_query = { SOI ~ query ~ EOI }
+query_prefix = { SOI ~ query }
```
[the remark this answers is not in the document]
flyingsilverfin: I want to be able to parse things that contain non-query text after, eg a query followed by `commit` or some junk.

rust/parser/typeql.pest
```diff
@@ -363,7 +361,7 @@ range_bound = _{ value_literal }
 
 // TYPEQL SYNTAX KEYWORDS ======================================================
 
-reserved = { WITH | MATCH | FETCH | UPDATE | DEFINE | UNDEFINE | REDEFINE | INSERT | PUT | DELETE
+reserved = { WITH | MATCH | FETCH | UPDATE | DEFINE | UNDEFINE | REDEFINE | INSERT | PUT | DELETE | END
```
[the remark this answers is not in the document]
flyingsilverfin: I tried that but this ends up not working:
```
define 
  entity person;
end;
```
The 'end' is interpreted as a label: 
```
query_define = { DEFINE ~ definables }

definables = { definable+ }
definable = { ( definition_type ~ SEMICOLON ) | definition_function | definition_struct }

// TYPE DEFINITION =============================================================

definition_type = { kind? ~ label ~ COMMA? ~ ( ( annotations | type_capability ) ~ ( COMMA ~ type_capability )* ~ COMMA? )? }
t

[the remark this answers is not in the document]
Comment (flyingsilverfin) [reply rust/parser/typeql.pest:364]: Reserving it feels right to me

File: rust/parser/typeql.pest (hunk 5 above)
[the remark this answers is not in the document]
Comment (flyingsilverfin) [reply rust/parser/typeql.pest:8]: good point, i think we can just query directly without SOI lol

File: rust/query/mod.rs (hunk 3 above)
Comment (flyingsilverfin) [reply rust/query/mod.rs:21]: Yea i loosely agree, except that it should be in `Query`:
```
query = { ( query_schema | query_pipeline_preambled ) ~ query_end? }
```

SchemaQuery is an enum over Define, Undefine, Redefine. Define is constructed in `visit_define_query`, which corresponds to this grammar path:
```
query = { ( query_schema | query_pipeline_preambled ) ~ query_end? }
query_schema = { query_define | query_undefine | query_redefine }
query_define = { DEFINE ~ definables }
```

Basically, we'd have to take in the 'end' in the ' query_define` rule, which doesn't feel right.

The correct thing to do would be to have:
```
struct Query {
  span: Option<Span>,
  structure: QueryStructure,
  explicit_end: bool
}

enum QueryStructure {
  SchemaQuery,
  Pipeline
}
```

which I'll probably just do now then

File: rust/parser/mod.rs [hunk 7]
```diff
@@ -126,24 +126,14 @@ fn parse_single(rule: Rule, string: &str) -> Result<Node<'_>> {
...
-        }
-        Err(error) => Err(error),
-    }
-}
-
 pub(crate) fn visit_eof_query(query: &str) -> Result<Query> {
     Ok(visit_query(parse_single(Rule::eof_query, query)?.into_children().consume_expected(Rule::query)))
 }
 
+pub(crate) fn visit_query_prefix(query: &str) -> Result<Query> {
```
flyingsilverfin, line 133: notably, not '_eof'!

rust/parser/typeql.pest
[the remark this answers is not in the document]
flyingsilverfin: Right so turns out that SOI does _something_? If i leave it out, then we get parsing errors. Just going to leave that in!

PR #406, opened 2025-04-26
Allow relates overrides to have lists

flyingsilverfin:

We fix one syntactic inconsistency, grammatically allowing list overrides for relation's roles:
```
define
  relation sub-rel, relates sub-role[] as super-role[];
```

## Implementation

We relax the grammar to allow specialisations of relation's roles with ordering.


farost (approved):
Lgtm

PR #412, opened 2025-08-26
Simplify type statements, allow empty define queries, bring back BDD

flyingsilverfin:

Composite changes:
1) We clean up the grammar for Type statements in both Patterns and Definables, which now allow commas after a `<kind> <variable>` query: 
```
match entity $x, sub $y;
```
is now legal with the comma. This is a more regular language that is easier to generate.

2) we allow empty define/undefine/redefine queries:
```
define
```

This addresses: https://github.com/typedb/typedb/issues/7531

3) We also re-enable half of our 2.x BDD suite, which parses every query in the behaviour repository (though we leave out the other 2.x half, which converts the parsed results back into strings, reparses, and validates the cycle is equivalent), ensuring we don't get parsing errors when we don't expect them and we do when they are expected.

## Implementation

- refactor grammar comma structure
- refactor definable to allow empty sequences
- re-build the BDD step implementations & update the BDD implementation to be a bit closer to what we have in Core


krishnangovindraj (changes requested):
Approved, but I'm hitting "request changes" just to flag the possible bug with printing out the schema.

PR #416, opened 2025-09-18
Add Quine trait to AST, make `pretty` public. 

sevki:

- feat:Make `pretty` public. 
- fix: Pretty now prints annotations. 
- fix: suppress some build warnings. 
- feat: add `derive(Quine)` to AST members. 

## Implementation

Adds an optional quine derive feature to the package that will implement [Quine](https://en.wikipedia.org/wiki/Quine_(computing))
for the AST to make it possible for the ASTs to reproduce them selves.


rust/variable.rs
```diff
@@ -6,15 +6,20 @@
 
 use std::fmt;
 
+#[cfg(feature = "quine")]
+use {polyquine::Quine, proc_macro2::TokenStream};
```
flyingsilverfin, line 10: Quine is a pretty old package - is it maintained?

PR #421, opened 2025-12-16
Allow trailing commas

flyingsilverfin:

To simplify writing and generating queries, we allow trailing commas anywhere we use a comma separated list of variables, statements, reductions, etc.


krishnangovindraj (approved):
Pure addition to the language.
I don't expect it to affect performance much but would still like a parsing benchmark implemented to for peace of mind.

flyingsilverfin:
Gonna put this into the xmas branch and merge it

PR #422, opened 2025-12-17
Use XID_START and XID_CONTINUE character classes

dmitrii-ubskii:

We use `XID_START` and `XID_CONTINUE` character classes provided by pest instead of manually listing unicode codepoint ranges for our identifiers. We also change the definition of the word boundary (`WB`) to be "any character that is not an identifier continuation character" that resolves parsing bugs such as `let $var= 4;` failing to parse: `=` is not `PUNCTUATION`, so not valid word boundary.

### DIFF
--- a/rust/parser/typeql.pest
+++ b/rust/parser/typeql.pest
@@ -588,11 +588,14 @@ ARROW = _{ "->" }
 
 // FRAGMENTS OF KEYWORDS =======================================================
 
-IDENTIFIER_LABEL_H = @{ XID_START }
-IDENTIFIER_LABEL_T = @{ XID_CONTINUE }
+IDENTIFIER_START = @{ XID_START }
+IDENTIFIER_CONTINUE = @{ "-" | XID_CONTINUE }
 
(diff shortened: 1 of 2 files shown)


rust/parser/typeql.pest
```diff
@@ -588,31 +588,14 @@ ARROW = _{ "->" }
...
-                   | '\u{FDF0}'..'\u{FFFD}'
-                   }
-IDENTIFIER_CONNECTOR = @{ "_"
-                        | "-"
-                        | "\u{00B7}"
-                        | '\u{0300}'..'\u{036F}'
-                        | '\u{203F}'..'\u{2040}'
-                        }
+IDENTIFIER_START = @{ XID_START }
+IDENTIFIER_CONTINUE = @{ "-" | XID_CONTINUE }
```
dmitrii-ubskii: `-` is not in XID_CONTINUE, but a valid continuation character in TypeQL identifiers.

flyingsilverfin (approved):

PR #432, opened 2026-02-09
Add tests for end; clause in query pipelines

flyingsilverfin:

Add tests for `end;` statements


krishnangovindraj (approved):
Looks good with one question

PR #446, opened 2026-04-28
@doc and @meta annotations

dmitrii-ubskii:

We implement `@doc("docstring")` and `@meta("key", "value")` schema annotations. These annotations can be attached to types and capabilities i.e. `owns`, `plays`, `relates`, and `sub`:

```php
define
entity person @doc("this represents an individual client") @meta("icon", ":silhouette.png"),
  owns name @doc("full name including title");
```
```php
match
  $x isa client;
  let $x_icon = get_meta("icon", $t); # get each client with a UI display icon hint
```

These annotations can also be attached to function definitions. They go at the end of the signature, before the colon introducing the body of the function:
```php
define
fun get_random_number() -> integer
    @doc("chosen by a fair dice roll"):
match
  let $rand = 4;
return first $rand;
```


rust/parser/undefine/mod.rs
```diff
@@ -82,23 +84,30 @@ fn visit_undefine_annotation_from_type(node: Node<'_>) -> AnnotationType {
...
-        Rule::ANNOTATION_REGEX => token::Annotation::Regex,
-        Rule::ANNOTATION_SUBKEY => token::Annotation::Subkey,
-        Rule::ANNOTATION_UNIQUE => token::Annotation::Unique,
-        Rule::ANNOTATION_VALUES => token::Annotation::Values,
-        _ => unreachable!("{}", TypeQLError::IllegalGrammar { input: child.as_str().to_owned() }),
-    }
+    let mut children = node.into_children();
+    let keyword = children.consume_any();
+    let annotation_category = match keyword.as_rule() {
+        Rule::ANNOTATION_ABSTRACT => AnnotationCategory::Abstract,
```
flyingsilverfin, line 92: Oh fun can we dissolve the core-specific IR for these now?

rust/parser/typeql.pest
```diff
@@ -220,7 +220,7 @@ relates_declaration = { RELATES ~ label_list ~ ( AS ~ label_list )?
 
 // FUNCTION DEFINITION =========================================================
 
-definition_function = { FUN ~ function_signature ~ COLON ~ function_block }
+definition_function = { FUN ~ function_signature ~ annotations? ~ COLON ~ function_block }
```
flyingsilverfin, line 225: so how do we expect to write this?
```
define
fun get_name($p: person) -> { name}
    @doc("..."):
  match $p has name $name;
  return { $name };
```

?

rust/schema/definable/function.rs
```diff
@@ -40,6 +48,9 @@ impl Pretty for Function {
         indent(indent_level, f)?;
         write!(f, "{} ", token::Keyword::Fun)?;
         Pretty::fmt(&self.signature, indent_level, f)?;
+        for annotation in &self.annotations {
+            write!(f, " {}", annotation)?;
```
flyingsilverfin: maybe we put the newline before each annotation with 4 spaces (2 more than a regular indent?)

flyingsilverfin (approved):
Approved w comments

rust/parser/undefine/mod.rs
dmitrii-ubskii: No, the `Meta` category carries a `StringLiteral` in typeql but a parsed unescaped `String` in core.

PR #456, opened 2026-07-10
Support vector value type

lolski:

Adds support for declaring `vector`. This lets users work with embeddings, and later express vector-similarity queries.

1. Declaring a vector embedding attribute 
```typeql
define
attribute embedding,
    value vector(64, "float32");
```

2. Inserting a vector embedding attribute: using declarative syntax
```typeql
insert
$x isa document,
    has embedding vector([1.0, 2.0, 3.0], "float32");
```

3. Inserting a vector embedding attribute: using `given` syntax
```typeql
given $e: vector(64, "float8");
insert
$x isa document,
    has embedding $e;
```

## Implementation

- Updated Pest grammar to allow array declarations
- AST & parser: `BuiltinValueTypeArray` struct and visitors
- Updated tests: declaring double array, inserting an array, and calling a function using an array

### DIFF
--- a/rust/parser/expression.rs
+++ b/rust/parser/expression.rs
@@ -11,6 +11,7 @@ use crate::{
     common::{Spanned, error::TypeQLError, token},
     expression::{
         BuiltinFunctionName, Expression, FunctionCall, FunctionName, List, ListIndex, ListIndexRange, Operation, Paren,
+        VectorLiteral,
     },
     parser::{type_::visit_label_scoped, visit_label},
     value::{Literal, StructLiteral, ValueLiteral},
@@ -78,10 +79,23 @@ fn visit_expression_base(node: Node<'_>) -> Expression {
         Rule::expression_function => Expression::Function(visit_expression_function(child)),
         Rule::expression_parenthesis => Expression::Paren(Box::new(visit_expression_parenthesis(child))),
         Rule::expression_list_index => Expression::ListIndex(Box::new(visit_expression_list_index(child))),
+        Rule::vector_literal => Expression::Vector(Box::new(visit_vector_literal(child))),
         Rule::label_scoped => Expression::ScopedLabel(visit_label_scoped(child)),
         Rule::label => Expression::Label(visit_label(child)),
         _ => unreachable!("{}", TypeQLError::IllegalGrammar { input: child.as_str().to_owned() }),
     }
+}
+
+fn visit_vector_literal(node: Node<'_>) -> VectorLiteral {
+    debug_assert_eq!(node.as_rule(), Rule::vector_literal);
+    let span = node.span();
+    let mut children = node.into_children();
+    children.skip_expected(Rule::VECTOR);
+    let list = visit_expression_list(children.consume_expected(Rule::expression_list));
+    let precision =
+        token::VectorPrecision::from(children.consume_expected(Rule::vector_precision).as_str().trim_matches('"'));
+    debug_assert_eq!(children.try_consume_any(), None);
+    VectorLiteral::new(span, list, precision)
 }
 
 fn visit_expression_list_index(node: Node<'_>) -> ListIndex {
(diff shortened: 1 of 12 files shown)


rust/parser/typeql.pest
```diff
@@ -169,6 +169,8 @@ list_range = { SQ_BRACKET_OPEN ~ expression_value ~ DOUBLE_DOT ~ expression_valu
 expression_struct = { CURLY_OPEN ~ struct_key ~ COLON ~ struct_value ~ CURLY_CLOSE }
 struct_value = { expression_value | expression_struct }
 
+vector_literal = { VECTOR ~ PAREN_OPEN ~ expression_list ~ COMMA ~ vector_precision ~ PAREN_CLOSE }
```
krishnangovindraj, line 172: If not function, why function shaped :(

flyingsilverfin: @krishnangovindraj  rust has the same:
```
// 1. Definition
struct Color(i32, i32, i32);
struct Point(f64, f64);
```

lolski: as a reader i don't find it confusing as it's easy to differentiate them based on the context. in schema declaration where a type is expected, seeing `vector(64, "float32")`  immediately tells me that 64 and "float32" is a type parameter of vector rather than arguments to a function in a function call.

rust/parser/expression.rs
```diff
@@ -78,12 +79,25 @@ fn visit_expression_base(node: Node<'_>) -> Expression {
...
         _ => unreachable!("{}", TypeQLError::IllegalGrammar { input: child.as_str().to_owned() }),
     }
 }
 
+fn visit_vector_literal(node: Node<'_>) -> VectorLiteral {
+    debug_assert_eq!(node.as_rule(), Rule::vector_literal);
+    let span = node.span();
+    let mut children = node.into_children();
+    children.skip_expected(Rule::VECTOR);
+    let list = visit_expression_list(children.consume_expected(Rule::expression_list));
```
flyingsilverfin, line 94: This will return a vector of expressions to the server yes? Do you want to do the validation there? I don't see any we wouldn't allow it, but just to be certain - you'll allow things like `vector([1+1, 5.0], "float32")` yes?

rust/parser/typeql.pest
krishnangovindraj: But this is the context for an expression rather than a type-declaration. It's not crazy to overload syntax, but this looks like it is just a built-in function named vector? And maybe it should be, because:
* It's not grouped with the other `value_literal`s, so it presumably can't be used in contexts where they can.
* The first argument is an `expression_list`, so it's going to be sent to the `ExpressionExecutor` all the time anyway.
* It's technically not a "literal" if it has to be evaluated.

rust/parser/typeql.pest
```diff
@@ -317,9 +319,11 @@ identifier_var = @{ IDENTIFIER_VAR_H ~ IDENTIFIER_VAR_T* ~ WB }
 
 // VALUES ========================================================
 
-value_type = { value_type_primitive | label }
+value_type = { value_type_vector | value_type_primitive | label }
 value_type_optional = { value_type ~ QUESTION }
 value_type_list = { value_type ~ SQ_BRACKET_OPEN ~ SQ_BRACKET_CLOSE }
+value_type_vector = { VECTOR ~ PAREN_OPEN ~ integer_literal ~ COMMA ~ vector_precision ~ PAREN_CLOSE }
+vector_precision = @{ "\"" ~ "float32" ~ "\"" }
```
dmitrii-ubskii, line 326: `"float32"` (etc. down the line) should be separate rules in the pest grammar, and `vector_precision` should be compound atomic `${ }`.
We already have the parser, we shouldn't need to trim its outputs.

rust/parser/typeql.pest
dmitrii-ubskii: A side effect of this is that nonsense like `vector([2024-12-12], "supercalifragilisticexpialidocious");` is not rejected by this grammar. It will fail to parse as a `vector_literal`, fall back to an `expression_function`, and the user gets an error saying `Could not resolve function with name 'vector'.` instead of a more useful error message like `Built-in function 'vector' cannot be applied to ...`.

If the vector literal constructor is function-like, I don't think it belongs in the grammar.

dmitrii-ubskii (changes requested):

PR #460, opened 2026-08-20
Allow scientific notation for double literal without decimal part

dmitrii-ubskii:

Allow literals like `1e-6` to parse without requiring `1.0e-6`.

## Implementation

Before: `Int Frac Exp?`
After: `Int (Frac? Exp | Frac)`


krishnangovindraj (approved):
The visitor already handles this?

dmitrii-ubskii:
> The visitor already handles this?

It's an atomic rule, the visitor does not need to handle changes to its internals. It is ultimately parsed into a value in core.

PR #463, opened 2026-09-03
Support fully qualified names for builtin functions to emulate namespacing

krishnangovindraj:
Allows function names to be of the form `first::second::third` and so on, allowing us to expand our set of functions without worrying about polluting the global namespace

### DIFF
--- a/rust/parser/typeql.pest
+++ b/rust/parser/typeql.pest
@@ -158,7 +158,8 @@ expression_list_index = { var ~ list_index }
 list_index = { SQ_BRACKET_OPEN ~ expression_value ~ SQ_BRACKET_CLOSE }
 
 expression_function = { expression_function_name ~ PAREN_OPEN ~ expression_arguments? ~ PAREN_CLOSE }
-expression_function_name = { builtin_func_name | identifier }
+expression_function_name = { namespaced_func_name | builtin_func_name | identifier }
+namespaced_func_name = { identifier ~ (NAMESPACE_SEP ~ identifier)+ } // Plain identifiers could be namespaced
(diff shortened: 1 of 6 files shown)


rust/parser/typeql.pest
```diff
@@ -536,6 +537,9 @@ FLOOR = @{ "floor" ~ WB }
 LEN = @{ "len" ~ WB }
 ROUND = @{ "round" ~ WB }
 
+// NAMESPACED FUNCTIONS
```
flyingsilverfin: Delete this comment, it's just the NAMESPACE delimiter, nothing to do with functions (just for now it's limited to that)

rust/parser/typeql.pest
```diff
@@ -158,7 +158,8 @@ expression_list_index = { var ~ list_index }
 list_index = { SQ_BRACKET_OPEN ~ expression_value ~ SQ_BRACKET_CLOSE }
 
 expression_function = { expression_function_name ~ PAREN_OPEN ~ expression_arguments? ~ PAREN_CLOSE }
-expression_function_name = { builtin_func_name | identifier }
+expression_function_name = { namespaced_func_name | builtin_func_name | identifier }
+namespaced_func_name = { identifier ~ (NAMESPACE_SEP ~ identifier)+ } // Plain identifiers could be namespaced
```
flyingsilverfin: do you want to generalize a bit further already so it's easier later:
1) create `namespaced_identifier` as a general rule
2) invoke it here

I think tbh we'll only want it to be here or in labels though - not in all identifiers (not labels at least).

flyingsilverfin (approved):
Remember, also needs bdd

PR #469, opened 2026-09-16
Introduce '?' short-circuit for expressions

krishnangovindraj:
Introduce '?' short-circuit operator for optional variables expressions


rust/parser/typeql.pest
```diff
@@ -314,6 +312,7 @@ iid_value = @{ "0x" ~ ASCII_HEX_DIGIT+ ~ WB }
 vars = { var ~ ( COMMA ~ var )* ~ COMMA? }
 
 var = ${ VAR_ANONYMOUS | var_named }
+var_or_optional = ${ var_optional | var }
 var_optional = ${ var ~ QUESTION }
```
flyingsilverfin, line 317: I think we replace this (see BDD comments) and call it `var_or_propagate`. And it's only usable in expressions and in fetch right now, wdyt?

